PR #2, opened 2016-09-02
Improve first time setup instructions

SimonJinaphant:
- Add high speed internet recommendation
- Add command to source the setup.bash file, for first time setup this wasn't done automatically.
- Update JetBrain step with additional info
- Use number list instead of bullet point list for setup step
- Improve UX with addition of echo statements to bash scripts; this will give the user some indication of what is happening, otherwise they won't know when the set up has completed.


SimonJinaphant:
Please Squash and merge to avoid polluting the commit history of master

garethellis0:
Thanks Simon!

PR #17, opened 2016-09-24
Vision port

ValRat:
Ported the code for vision that we will most likely use for this year.


emsal1863:
file is currently named `ros-vision-test.cpp.cpp`, might want to change

garethellis0:
Just to confirm I understand the parameter server correctly, the parameters used are node-specific since you access them via` nh_private.param`?

ValRat:
Yeah, I'll just leave them as private. Not too hard to switch it to public if need be.

PR #22, opened 2016-10-16
Implemented CI (Continuous Integration)

garethellis0:
- Uses Travis CI
- Since Travis CI does not yet support Ubuntu 16.04 (required for ROS Kinetic) all tests are run in a docker image within Travis CI
- Build pass/fail image in readme corresponds to the status of master
- Uses the script located in [Snowbots fork of move_ci](https://github.com/UBC-Snowbots/moveit_ci)


garethellis0:
Should also note, full check takes around 8-9 minutes currently, majority of the time is setting up docker, with another significant chunk for installing dependencies.

garethellis0:
...... just ignore the failing check...... it's trying to pull the dependency list (.rosinstall) from the master branch (where it's being merged into), but it doesn't yet exist in master (it's part of this pull request). Should be fine in the future.

PR #24, opened 2016-10-21
Begin process for turning the src/firmware directory into an arduino …

emsal1863:
…workspace

The goal of this change is to make the `src/firmware` directory of the project into an actual arduino workspace, so developers can point their IDEs at it as their "sketchbook" directory and can just work straight out of it without having to worry about installing libraries.

(This needs some more work and testing but this PR wasn't created when I had actual access to an Arduino, I'll test it tonight)


garethellis0:
Sounds good Em, just comment on this PR once you've tested it and I'll merge it!

emsal1863:
done

PR #26, opened 2016-10-21
Readme fixes

ValRat:


garethellis0:
Pft, not even any useful code. Typical Val.

PR #31, opened 2016-10-27
Usb detection

emsal1863:
this being created so other people can see as this progresses. not meant to be merged until i edit this message


src/drivers/src/USBDetection.cpp
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        for (dhandle_t &it : handles) {
+            
+            libusb_device_handle *current_handle = it.get();
+
+            int numBytes;
+            unsigned char data[512];
+
+            int transfer = libusb_bulk_transfer(current_handle, LIBUSB_ENDPOINT_IN, data, sizeof(data), &numBytes, 0);
```
fhackett: This line confused me a bit - you could add a comment explaining what this does (LIBUSB_ENDPOINT_IN for example - I had to look that up).

EDIT: see my comment about the read function - reading USB might be more complicated than this suggests.

src/drivers/src/USBDetection.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+            std::string dataStr((char *)data, 0, 10);
+            if (dataStr.compare(0, prefix.length(), prefix) == 0) {
+                return ArduinoConnection(std::move(it));
+            }
+        }
+        return ArduinoConnection();
+    }
+
+    ArduinoConnection UsbDetector::open_arduino_with_prefix(char *prefix) {
```
fhackett: Is this function necessary? I think `const char*` autoconverts to std::string if needed, so you can call the std::string version with "prefix" and it should work anyway...

src/drivers/src/USBDetection.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::vector<dhandle_t> handles(sz);
+        
+        int i = 0;
+        for (usbdev_t const &it : arduinos) {
+            libusb_device_handle *devhandle;
+            int err = libusb_open(it.get(), &devhandle);
+            if (err == 0) {
+                handles[i] = dhandle_t(devhandle, libusb_close);
+            } else {} //TODO
+            i++;
```
fhackett: This confuses me... how can you increment something that's const? The range-for should handle all that for you? (pls correct me if I misunderstood)

src/drivers/src/USBDetection.cpp
```diff
@@ -0,0 +1,96 @@
...
+        int numBytes;
+        int transfer = libusb_bulk_transfer(arduino, LIBUSB_ENDPOINT_OUT, data, data_sz, &numBytes, 0);
+    }*/
+
+    std::string ArduinoConnection::read() {
+        unsigned char data[512];
+        int numBytes;
+        int transfer = libusb_bulk_transfer(arduino.get(), LIBUSB_ENDPOINT_IN, data, sizeof(data), &numBytes, 0);
+
+        return std::string(data, data+numBytes);
```
fhackett: Tricky note 1: this may not be the most useful return type. What if the user wants to read raw binary data that is not text? Unsigned-signed conversion will mess this up + the result will not be a meaningful string.
You could return a `std::vector<unsigned char>` to fix this issue.

Tricky note 2: the user can't ask for more or less than 512 bytes... this might work often, but in more complex cases will stop working correctly (you get get half a message or maybe no message at all, etc...).

This seems to be a complex issue, see http://www.beyondlogic.org/usbnutshell/usb4.shtml - do we really want to use bulk transfer here anyway? What kind of information do we expect to read? If you've already discussed this, then adding a comment may help future developers (and me). If not, this is the part where we think carefully about how we want to transfer data and what kind of data we want to transfer.

src/drivers/src/USBDetection.cpp
garethellis0: I don't think `i` is const here, are you thinking of `it`?

src/drivers/src/USBDetection.cpp
garethellis0: I think bulk transfer should be appropriate for the current system. We should only be passing things like accelerometer, gyroscope data, nothing like video/audio/lidar streams. Were we to move to having this control everything, using Isochornous transfers for those sorts of things would probably be appropriate, perhaps something to think about for the future?

garethellis0: Also, I second the remark about a detailed comment being needed here.

src/drivers/src/USBDetection.cpp
emsal1863: `i` is just an `int` being used as a counter value for the number of arduinos that can be opened without error. Perhaps I should change the variable name so it's less confusing.

I ran into some problems with insufficient permissions errors when I don't run a program that uses this call as root, manual testing that.

src/drivers/src/USBDetection.cpp
emsal1863: ah, nice. Removed

src/drivers/src/USBDetection.cpp
emsal1863: I'll change the return type and add a comment about this stuff, not totally sure about bulk transfer.

I'm trying to manually test reading arduino `Serial.write()` data using libusb bulk transfers, and so far I'm getting incorrect output. I'll be investigating further.

src/drivers/src/USBDetection.cpp
fhackett: It seems I misread... never mind, this works. As you guessed, brain replaced `i` with `it`.

fhackett: @emsal1863 Regarding root permissions, you can try to get your user to be a member of the dialout group... permissions :D

src/drivers/src/USBDetection.cpp
fhackett: @garethellis0 Isn't accelerometer/gyroscope streaming as well? Not sure why we're picking bulk when using a different priority would be just a function name change away... does using another mode complicate things? ...maybe the arduino is missing some USB functions?

@emsal1863 cool on the comment and return type - that should work.

Two more changes I can think of:
- (if you haven't already) documenting that `read()` might return short reads (you can check the length of the result to see what you got), so users don't forget that
- allowing a template parameter for expected read size. 512 is kind of random, and I feel that having it as a changeable default rather than hard-coded would be more extensible (maybe we know exactly how long the message should be in some cases).

src/drivers/src/USBDetection.cpp
garethellis0: @fhackett oh god.... this again :grimace: :stuck_out_tongue_closed_eyes:

src/drivers/src/USBDetection.cpp
garethellis0: @fhackett what function were you thinking of using? Looking at the [libusb doc](http://libusb.org/static/api-1.0/io.html) on synchronous and asynchronous transfers, the only synchronous transfer is libusb_bulk_transfer. 

While I agree that using an asynchronous transfer is probably the best for our purposes, it states in the doc that "it is a more complex interface" (and looking at some examples seems to support this). I'm mostly trying to avoid any asynchrony not internal to the class, as it will likely be used in the future by **extremely** inexperienced devs, who may not be familiar with it.

I only have a very rudimentary understanding of this, and I'm likely misunderstanding some things, so if you have a second, could you perhaps do a branch of this branch and write a rough implementation of what you're thinking of? No pressure to do so, it would just help me understand a bit. 

Otherwise, going forward, there are a few options, and seeing as Em's responsible for this, it's totally up to him what he wants to do. (That being said, the Jan 15th deadline for a rough working product is still in place)

a) re-write this to use libusb's asynchronous interface
b) use bulk_transfer for now, then make (a) an issue to be addressed in another PR

emsal1863: For some reason, bulk transfers aren't producing the correct output (it isn't giving whatever the arduino was writing via `Serial.write`)

Investigating further. Perhaps `libusb` isn't necessarily the best library for communication with a serial point -- pretty annoying.

Also, I support option `b`: doing all of the async stuff later

garethellis0: Hmm, sounds good, you should have permissions to create issues, let me know if you can't

src/drivers/include/USBDetection.h
```diff
@@ -21,10 +21,10 @@ namespace USBDetection {
...
             ArduinoConnection() = default;
             ArduinoConnection(dhandle_t);
-            std::string read();
+            std::vector<unsigned char> read();
             template <typename deserializable> void write(const deserializable &obj) {
                 auto msg = obj.serialize();
-                
+
                 int numBytes;
                 int transfer = libusb_bulk_transfer(arduino.get(), LIBUSB_ENDPOINT_OUT, msg.data(), msg.size(), &numBytes, 0);
```
fhackett: one minor comment here: what if there's an error? We seem to just assume transfer will be 0... if this is actually valid for our purposes, maybe use the assert macro? assert(transfer == 0) (included in cassert header)

(same goes for the read function)

src/drivers/include/USBDetection.h
```diff
@@ -21,10 +21,10 @@ namespace USBDetection {
         public:
             ArduinoConnection() = default;
             ArduinoConnection(dhandle_t);
-            std::string read();
+            std::vector<unsigned char> read();
             template <typename deserializable> void write(const deserializable &obj) {
```
fhackett: Oh hey cool I didn't notice you'd written this :)

src/drivers/src/USBDetection.cpp
```diff
@@ -86,11 +82,15 @@ namespace USBDetection {
         int transfer = libusb_bulk_transfer(arduino, LIBUSB_ENDPOINT_OUT, data, data_sz, &numBytes, 0);
     }*/
 
-    std::string ArduinoConnection::read() {
+    std::vector<unsigned char> ArduinoConnection::read() {
```
fhackett: This should work :)

garethellis0:
@emsal1863 if this is no longer being developed (as indicated by you starting a new branch) could you please close it

emsal1863:
dropped in favour of serial detection

ValRat:
Seems to not be developed anymore, deleting branch.

PR #40, opened 2016-10-30
Gps decision

chrischenubc:


src/decision/src/GpsDecision.cpp
```diff
@@ -17,19 +19,128 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
+         }
+     }
+    if(x==0 && y==0 && current_heading==0) {
+        desiredAngle==0;
+    }
+
+    return desiredAngle;
+}
+
+void  GpsDecision::rotate(double desiredAngle,double angular_velocity) {
```
garethellis0: According to my IDE, this function is never called?

src/decision/include/GpsDecision.h
```diff
@@ -14,15 +14,52 @@
...
     GpsDecision(int argc, char **argv, std::string node_name);
+
+    /**caclutes the distance between wayPoint and current point
+     * @ parameter: given the gps locaiton of the next move
+     * @ return: the distance between the next location and the current location
+    */
+    static double distance(const geometry_msgs::Point::ConstPtr& relative_gps);
+
+    /**
+ * @ parameter
```
garethellis0: Indenting should be fixed

src/decision/src/GpsDecision.cpp
```diff
@@ -17,19 +19,128 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
+    double x=relative_gps->x;
+    double y=relative_gps->y;
+    double z=relative_gps->z;
+
+    double distance=sqrt(pow((x-current_x),2)+pow((y-current_y),2)+pow((z-current_z),2));
+
+  return distance;
+}
+
+#define TO_DEGREES 180/(M_PI)
```
garethellis0: this should be declared as a private `const float` in the header, and initialised in the constructor.

src/decision/src/GpsDecision.cpp
garethellis0: Please either delete this function if it is not used, or provide a comment explaining why it's here.

garethellis0:
Hey @chrischenubc , you seem to have some failing tests. Before you go about improving the node, do you think you could fix them? Seems like your code isn't compiling properly? Have you tried running catkin_make on your machine?

src/decision/include/GpsDecision.h
```diff
@@ -14,15 +14,51 @@
...
+    /**caclutes the distance between wayPoint and current point
+     * @ parameter: given the gps locaiton of the next move
+     * @ return: the distance between the next location and the current location
+     **/
+    static double distance(const geometry_msgs::Point::ConstPtr& relative_gps);
+
+    /**
+     * @ parameter
+     * relative_gps: given the gps locaiton of the next move
+     * current_heading: the curren heading relative to north in degrees (0 to 360 degrees)
```
garethellis0: Also, it's fine for this code, but generally we use radians rather then degrees, just for future reference.

src/decision/include/GpsDecision.h
```diff
@@ -14,15 +14,51 @@
...
+     * @ return:
+     *   desiredAngle: the angle(less than PI) in degrees that robot need to turn
+     *   positive value means turning clockwise
+     *   negative value means turing counter-clockwise
```
garethellis0: I believe that the variable and initial description should be on the same line. Also, you don't need a variable name for the return value unless it actually adds meaning to the documentation. So this would be more like
```
@return the angle(less than PI) in degrees that robot need to turn
positive value means turning clockwise
negative value means turing counter-clockwise
```

src/decision/include/GpsDecision.h
```diff
@@ -14,15 +14,51 @@
...
+     */
+    static double desiredAngle(const geometry_msgs::Point relative_gps,float current_heading,geometry_msgs::Point currentPoint);
+    /**
+     * @ parameter
+     *  desiredAngle: the angle(less than PI) in degrees that robot need to turn
+     *                 positive value means turning clockwise
+     *                 negative value means turing counter-clockwise
+     *  angular_velocity: the angular_velocity(rad/s)
+     * @ effects: rotate the robot by giving angles and angular velocity
```
garethellis0: is `@effects` a valid flag? I've never seen it before. (It might be, I'm just asking to make sure)

garethellis0: Again, general formatting, see above comments.

src/decision/include/GpsDecision.h
```diff
@@ -14,15 +14,51 @@
...
+    private:
+    //gps callback for the current point
+    void gpsCurrentCallBack(const geometry_msgs::Point::ConstPtr& relative_gps);
+    //gps callback for the next point
     void gpsCallBack(const geometry_msgs::Point::ConstPtr& relative_gps);
+    //gps callback for the current heading
+    void compassCallBack(const std_msgs::Float32::ConstPtr& compass_heading);
     void publishTwist(geometry_msgs::Twist twist);
-
+    ros::Subscriber compass_subcriber;
```
garethellis0: Spelling

src/decision/src/#gps_decision.cpp#
```diff
@@ -0,0 +1,24 @@
+/*
+ * Created By: YOUR NAME HERE
```
garethellis0: Give yourself some credit! :stuck_out_tongue:

src/decision/src/#gps_decision.cpp#
```diff
@@ -0,0 +1,24 @@
+/*
```
garethellis0: I don't think this file should exist, looking at the file name it looks like a temp file.

src/decision/src/GpsDecision.cpp
```diff
@@ -17,19 +19,133 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
 
     // Setup Subscriber(s)
     std::string gps_filtered_topic_name = "/gps_conversion/relative_gps";
+    std::string compass_topic_name = "/gps_conversion/compass";
     int refresh_rate = 10;
+    gps_subscriber = public_nh.subscribe(gps_filtered_topic_name, refresh_rate, &GpsDecision::gpsCurrentCallBack, this);
     gps_subscriber = public_nh.subscribe(gps_filtered_topic_name, refresh_rate, &GpsDecision::gpsCallBack, this);
-
+    compass_subcriber = public_nh.subscribe(compass_topic_name, refresh_rate, &GpsDecision::compassCallBack, this);
```
garethellis0: spelling

src/decision/src/GpsDecision.cpp
```diff
@@ -17,19 +19,133 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
+    const double TO_RADIANS=(M_PI)/180;
+
+    double next_x=relative_gps.x;
+    double next_y=relative_gps.y;
+    //calculate the relative x and y
+    double x=(next_x-currentPoint.x);
+    double y=(next_y-currentPoint.y);
+
+    //double dis=distance(relative_gps);
+    double tan=y/x;
```
garethellis0: avoid naming variables like "tan", "sin", "cos", "abs", etc. as this may cause confusion and/or naming conflicts since there are mathematical functions with the same names

src/decision/src/GpsDecision.cpp
```diff
@@ -17,19 +19,133 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
+    do{
+        //velocity_publisher.publish(vel_msg);
+        double t1=ros::Time::now().toSec();
+        current_angle=angular_velocity*(t1-t0);
+        ros::spinOnce();
+        loop_rate.sleep();
+        //loop_rate
+    }while(current_angle<desiredAngle);
+    vel_msg.angular.z=0;
+    // velocity_publisher.publish(vel_msg);
```
garethellis0: Commented out code should be removed. That being said, I think you need this line to publish the twist message you create.

src/decision/test/gps-decision-test.cpp
```diff
@@ -1,13 +1,41 @@
...
+    mypoint.x = 6;
+    mypoint.y = 3;
+    mypoint.z = 0;
+    double expected=sqrt(pow(mypoint.x,2)+pow(mypoint.y,2)+pow(mypoint.z,2));
+    geometry_msgs::Point::ConstPtr mypointer(new geometry_msgs::Point(mypoint));
+    EXPECT_DOUBLE_EQ(expected, GpsDecision::distance(mypointer));
+}
+
+TEST(GpsDecision,desiredAngle) {
+    #define PI acos(-1.0)
```
garethellis0: Don't think you need this, should just me able to use `M_PI` as you have elsewhere in the code (I think)

src/decision/test/gps-decision-test.cpp
```diff
@@ -1,13 +1,41 @@
...
+    currentPoint.x = 3;
+    currentPoint.y = 3;
+    currentPoint.z = 0;
+
+    float my_heading=15;
 
+    double expected=(-150*M_PI/180);
+
+    EXPECT_DOUBLE_EQ(expected, GpsDecision::desiredAngle(nextpoint,my_heading,currentPoint));
+
```
garethellis0: This test is good, but I'd like to see some more, try testing with the relative point in each of the 4 quadrants would be a good start.

garethellis0:
Old gps_decision stuff. Should be redone soon and a new pr opened.

ValRat:
Code been merged in elsewhere, deleting branch.

PR #72, opened 2017-01-28
added sick packges, should have no need for dependencies

Dulluhan:
Added sick packages as submodules


garethellis0:
This looks good, presuming that it runs fine on your system. Only thing is that if we're going to include submodules in the repo, the installation instructions will need to be changed accordingly. Basically, instead of `git clone`, we'll have to `git clone --recursive` on initial setup. Just make the change to the ReadMe in the setup section and we should be good!

PR #79, opened 2017-02-05
Lidar Decision Node

### LINKED ISSUE #46: Lidar Decision

garethellis0:
- Currently, the node is pretty simple, especially in regards to how the twist message is created (right now it just turns when near an obstacle)
- closes #46


src/decision/src/LidarDecision.cpp
```diff
@@ -15,22 +16,115 @@ LidarDecision::LidarDecision(int argc, char **argv, std::string node_name) {
...
+    }
+}
+
+geometry_msgs::Twist LidarDecision::twist_message_from_obstacle(LidarObstacle obstacle,
+                                                               distance_t danger_distance,
+                                                               angle_t danger_angle,
+                                                               float linear_vel_multiplier,
+                                                               float angular_vel_multiplier) {
+    geometry_msgs::Twist twist;
+    twist.linear.x = 0;
```
ValRat: Would this never make the robot go forward?

src/decision/test/lidar_decision_rostest.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // We would also not want to be going forward
+    EXPECT_EQ(0, command.linear.x);
+
+    // Everything else should always be 0
+    EXPECT_EQ(0, command.linear.y);
+    EXPECT_EQ(0, command.linear.z);
+    EXPECT_EQ(0, command.angular.x);
+    EXPECT_EQ(0, command.angular.y);
+}
+
```
ValRat: Maybe add a test case where there's no obstacles? See how it handles zero input.

src/decision/src/LidarDecision.cpp
garethellis0: I'm presuming that an all 0 message will be treated by the central decision node as no input (ie. it will either use other decision nodes or just go forward)

garethellis0:
@ValRat another review plz

PR #91, opened 2017-02-12
Setup script fixes

### LINKED ISSUE #69: Write a new setup script
- Current one is a bit of a hacky mess in bash
- Does not need to be bash, other frameworks/programs are totally an option as well
- Needs to add things to `.bashrc` and install programs, via `apt-get` for stuff like ROS, and by manually wget'ing and install other packages, such as Clion

### LINKED ISSUE #78: Fix Gazebo environment variable
Gazebo does not know where to find the world files. So add it in the environment. (in .bashrc).
Best to do this by amending the get_started.sh script

export GAZEBO_MODEL_PATH=/home/path/to/IGVC-2017/src/sb_gazebo/models:${GAZEBO_MODEL_PATH}
export GAZEBO_PLUGIN_PATH=/home/path/to/IGVC-2017/src/sb_gazebo/lib:${GAZEBO_PLUGIN_PATH}
export GAZEBO_RESOURCE_PATH=/home/path/to/IGVC-2017/src/sb_gazebo/models:${GAZEBO_RESOURCE_PATH}

garethellis0:
- Simplified how we write the new lines to `bashrc` and `zshrc` files,
  as well as making it easier to add more file/lines to add
- Changed the install location of CLion to a more sensible one
- Updated the CLion version

Closes #69, closes #78

### DIFF
--- a/get_started.sh
+++ b/get_started.sh
@@ -6,15 +6,59 @@ echo "Starting first time installation and setup, please wait"
 echo "these downloads can take a while if you're on slow internet."
 echo "================================================================"
 
+# The current directory
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-# Shell config files to add aliases to
-SHELL_CONFIG_FILES=("$HOME/.bashrc" "$HOME/.zshrc")
 
-###############################
-# Install ROS
-###############################
+echo "================================================================"
+echo "Setting up your shell config files"
+echo "================================================================"
+# Shell config files that various shells source when they run.
+# This is where we want to add aliases, source ROS environment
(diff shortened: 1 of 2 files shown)


ValRat:
So if I ran this again, it would install another clion in a different directory? If so might be worth to toss a warning (at readme, verbally) at older members if this is the case.

garethellis0:
It will install to a different directory, but as far as I could tell all settings are preserved. I opened CLion after running it and even had the same files open. Since I don't think people really pay attention to where it gets installed, unless you can think of other side effects, I don't think a note is needed.

ValRat:
Sounds fine then, merging. :+1:

PR #97, opened 2017-02-18
Serial detection

emsal1863:
Still mulling over some of the design details of this class (might want to make it , but here is a naive synchronous implementation of the functionality.


src/drivers/src/SerialDetection.cpp
```diff
@@ -0,0 +1,42 @@
...
+    if ((dir = opendir("/dev/serial/by-id/")) != NULL) {
+        while((ent = readdir(dir)) != NULL) {
+            possibleSerialFilenames.push_back(std::string(ent->d_name));
+        }
+    } else {
+        throw direx;
+    }
+}
+
+// return an fstream of the arduino broadcasting the appropriate data
```
garethellis0: Just generally, this comment should probably go in the header documentation I mentioned above.

garethellis0:
Some magic udev rules discovered by @lujinhao pretty much remove the need for this (at least for our current purposes)

PR #98, opened 2017-02-18
add Zed Configuration and ros node

lujinhao:


garethellis0:
**Documentation of evening's discussion:**
- We will not be able to compile the zed repo,  as it requires CUDA, and fat bloody chance that we can get that to run in CI :laughing: 
- So we need the ability to blacklist certain projects from being built when CI builds the entire system (because it will fail if it tries to build the zed project)
- We have decided to switch to `catkin_tools`, the replacement for the old catkin build system. It's cutting edge :hocho: , in beta, and of questionable stability.... just how we like it.
- As such, some commands have changed (as reflected in the updated readme). There is also a new alias (namely two lines that need to be added to `.bashrc`) that can be added by running `get_started.sh` again, or just adding these lines to your `~/.bashrc`:
```
    alias catkin='catkin '                                                                              
    alias run_tests='run_tests && catkin_test_results'  
```
- CI has already been updated to use the new build commands.
- Generally speaking, using the old commands should still work just fine, we've just had to migrate CI to the new commands so we can use the new blacklist functionality, and for consistency we'd like to migrate everything over to the new commands eventually, because it's new and shiny (seriously, building is bloody beautiful now, so colourful)

**Before we merge this PR**
- need to fork the zed repo and update the `opencv` requirement in the package.xml to use `opencv3` so rosdeps can find and install it. We should also open a PR on the original zed repo asking to merge our changes back so we don't have to maintain our own fork. Either way, the zed will still be a submodule on the main repo.
- Notify people of the changes to build commands

@ValRat feel free to add anything I missed here

PR #104, opened 2017-02-25
Fixed sb_gazebo nodes not being compile

garethellis0:
If `catkin_package()` is not in the CMakeLists.txt, it causes issues, such as nodes not being compiled. Re-added to sb_gazebo.


garethellis0:
Pushing this through, as it is a one line revert, and literally breaks gazebo.

PR #106, opened 2017-02-25
Vision stack cleanup

### LINKED ISSUE #35: Save Color Calibration to Param Server
Right now the params are being saved to a file in a mildly sketchy location. They should be saved on the param server. For more details ask @ValRat

### LINKED ISSUE #73: Fix vision IPM coordinate settings
Make all default x and y values relative to width and height.

### LINKED ISSUE #96: No Windows if no camera input for vision stack
In the vision stack (currently the `ros_vision` node) if the node does not receive any camera input, then no windows will open up (normally there is one for raw camera, one for filtered camera, one with IPM, and the final image). The windows should still pop up, but display a message saying something like "No camera input". Alternatively (and much more simply) a `ROS_WARN` message could be added to the node with a similar warning.

garethellis0:
- Put the 3 windows showing the various stages of the image translation/filtering into a single window, made the height and width window params
- Updated RosVision so that we now wait for our first image before checking for our params for the IPM and creating the IPM. This allows us to set sensible default values for all params related to the IPM (like the points we map from and the points we map to)
- Changed where `filter_init.txt` is saved to, it's now saved in `src/vision/launch/filter_init.txt`, which allows us to always find it, no matter where we run the node from

**Still need to fix merge conflicts**

@ValRat - could you please test this throughly, especially all the new params I added, and make sure that they act as you ~behave~ expect (also to test my naming)

Closes #96, closes #73, closes #35


src/sb_utils/include/utils.h
```diff
@@ -0,0 +1,40 @@
+/*
```
garethellis0: Yeah, removed

garethellis0:
@ValRat - requested changes made

src/vision/src/RosVision.cpp
```diff
@@ -91,15 +116,25 @@ void RosVision::imageCallback(const sensor_msgs::ImageConstPtr &msg) {
...
-    namedWindow(ipmOutputWindow, CV_WINDOW_AUTOSIZE);
-    namedWindow(filterOutputWindow, CV_WINDOW_AUTOSIZE);
-}
-
-void RosVision::deleteWindow() {
-    destroyWindow(inputWindow);
-    destroyWindow(ipmOutputWindow);
-    destroyWindow(filterOutputWindow);
+    // Create one big mat for all our images
+    cv::Size main_window_size(displayWindowWidth, displayWindowHeight);
```
ValRat: displayWindowWidth and displayWindowHeight doesn't determine the actual overview window size, just the size of the Mat which holds the image collage. Since we create the window with the CV_WINDOW_NORMAL flag, which allows for size changes (compared to CV_WINDOW_AUTOSIZE which makes the window the size of the underlying Mat with no window resize). I think we can deprecate these two parameters and have them autoset by the size of the image.

ValRat:
Also maybe add this to Readme.md within this package or something: 

```
Image Parameters
width (int) - input image width
height (int) - input image height

IPM Bounding Box Parameters
0 <= value <= 1
ipm_base_width (double) - the length of the bottom of the box as % of image
ipm_top_width (double) - the length of the top of the box as % of image 
ipm_base_displacement (double) - displacement of the box from the bottom of the image 
ipm_top_displacement (double) - displacement of the box from the top of the image

update_frequency (int) = fps update
display_window_width (int) - the width of the overview window
display_window_height (int) - the height of the overview window
config_file (string, absolute path) - the location of the config file
show_image_window (boolean) - displays the overview window (Can be toggled by ‘s’)
show_calibration_window (boolean) - displays the calibration window (Can be toggled with ‘m’)
```

src/vision/README.md
```diff
@@ -0,0 +1,25 @@
+# Vision
+- This package contains everything pertaining to the vision stack that we develop in-team (ie. no external code). 
+- This includes nodes for filtering and basic camera launch nodes.
+- This package may also contain launch files for launch both our nodes, and any external nodes related to vision
+
+# Nodes
+## ros_vision
+- This is our IPM (Inverse Prespective Mapping - basically translating things from the camera's po**int** of view to a birds-eye po**int** of view) and HSV (Hue, Saturation, Value - used for filtering out white lines) node.
```
ValRat: *Perspective

ValRat: and I think your search and replace had some un**int**ended side effects

garethellis0: :rofl:

PR #112, opened 2017-03-01
Implemented a better dangerScore function

raadkhan:
- Calculated danger score by taking sum of obstacle's average angle score and obstacle's closest distance score (relative to the robot)
- Wrote two supplementary functions: getMinDistance( ) and getMaxDistance( )
- Wrote a gtest for dangerScore( ) and  passed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,10 +10,7 @@ clion*
 .swp
 filter_init.txt
 cmake-build-debug
-<<<<<<< HEAD
 
 # We want to ignore the catkin settings, except for some specific files
 /.catkin_tools
 !/.catkin_tools/profiles/default/config.yaml
-=======
->>>>>>> 05febcf269ce3cda133c784c333f29b2af39bf51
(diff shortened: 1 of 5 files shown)


raadkhan:
Forgot to push LidarDecision.h which contained the new function definitions, and need to resolve merge conflicts

raadkhan:
Pushed LidarDecision.h with new function definitions

garethellis0:
Everything looks good, but git appears to have sh** the bed, and thinks you deleted and re-added massive chunks of the file. Before I merge, can I confirm that all you've done is added `getMinDistance()` and `getMaxDistance()`, updated the `dangerScore()` function, and added a test for it?

Also, you can put stuff in code boxes in git comments like this: \` some code \`, or if you want to put it in a block, like this:
\```
Here is some code
\```

(I've escaped the \` characters above so you can see them. Normally this would look like this: `some code` and this:
```
Here is some code
```

raadkhan:
I can confirm I have not deleted any files.
```
Here is some more code
```

garethellis0:
Yeah, I was more referring to the changes you made, namely:
- added getMinDistance() and getMaxDistance()
- updated the dangerScore() function, and added a test for it

raadkhan:
Ah right and made the minor fixes, yes

PR #122, opened 2017-03-09
Created a GPS Driver and Cleaned up some stuff

### LINKED ISSUE #102: Write a ROS driver for our GPS Arduino
- The Arduino firmware for the GPS is complete and working, but we need a ROS node to interface with it. 
- The node would need to get the gps readings sent by an arduino running `firmware/gps_firmware/gps_firmware.ino`. 
- This means also checking for "no signal" (or whatever the equivelent is that the gps firmware sends) and not publishing a ROS message in this case (or some other sort of indication that we do not have a signal, perhaps a boolean on a separate topic).

garethellis0:
- Created a GPS driver to get info from `gps_firmware.ino`
- Updated the `GPSManager` to get the compass heading from the razor_imu (it has to extract it from the Quaternion representing orientation now)
- Cleaned up the firmware project
- Closes #102 
@ValRat - Before you approve this, please get me to give you the GPS ASAP and make sure that everything runs on your system


ValRat:
I like, just need testing when GPS signals aren't being dumb

Issue #101, opened 2017-02-19
Warning Error being thrown when tests are run


garethellis0:
There is an `ROS_WARN` being thrown when the tests are run, namely:
```
[ERROR] [1487482866.179100462]: Given odd length list of waypoints, check that "waypoints" parameter has an (numerically) even length. As this is a series of lat/lon coordinates,it should always be even
```
This is an expected error, originating from the fact that the test causing it is actually testing an odd number of points. This error is causing warnings in the tests, and so should somehow be removed, or modified so that it indicates that it is acceptable for this to fail in testing (ie. add `[t is acceptable for this to fail in testing]` to the end of it, or something like that)

raadkhan:
Just to clarify, am I supposed remove the test failures simply by modifying ```EXPECT_EQ(0, command.linear.x);``` such that it evaluates as true and then comment ```it is acceptable for this to fail in testing```?

raadkhan:
@garethellis0 I have been attempting to actually fix the test issues by first finding where the error is coming from. I did not notice anything odd except in ```oneObstacleStraightAheadTest``` ```obstacleToLeft``` and ```obstacleToLeft``` all call ```std::fill( )``` and are these are the three tests which fail, but ```noObstacles``` does not call ```std::fill( )``` and is the only test that passes. Specifically, I see that changing the value to be populated in the range (third parameter) changes ```command.linear.x``` and command.linear.z but I cannot a see a relation... so how to make the assigned value such that ```command.linear.x``` evaluates to = zero and ```command.linear.z``` evaluates to >= 1e-5

garethellis0:
@raadkhan - this issue has doesn't have anything to do with the problem you're experiencing, so I'm going to move discussion over the the PR.

PR #128, opened 2017-03-17
Better way of getting a twist message from an obstacle

raadkhan:
The build summary states 3 failures though I only find this error under catkin run_tests

Job #539.1 line 1349 and job #539.2 line 1609: [ERROR] [1489766663.630440745]: Given odd length list of waypoints, check that "waypoints" parameter has an (numerically) even length. As this is a series of lat/lon coordinates,it should always be even

this occurs while GpsManager tests are running, though all 17 tests pass.


garethellis0:
The error you noted above is actually totally fine. That particular test is testing to **make sure** that a function fails, and so forces the "Error" message you see. It's not part of the test errors.... (but we should probably figure out how to fix it...... I've had an issue open for a while: #101). The actual failures here are some "Rostest's", which are basically tests the node by publishing some messages to a topic that the node subscribes to, and seeing what the node does (in this case, what it publishes). This particular rostest can be found in `src/decision/test/lidar_decision_rostest.cpp`. You can run the tests by first building (`catkin build`), then running the tests with `catkin run_tests` (unfortunately, this runs all the tests, we're working on a better solution: #130). 

You've unfortunately made all your changes on a branch which already existed, and had changes on it that we **don't** want to merge. To fix this, you will need to "cherry-pick" your changes over to a new branch. Cherry picking is where you take specific commits from one branch, and apply them to another. There is a good explanation (here)[http://think-like-a-git.net/sections/rebase-from-the-ground-up/cherry-picking-explained.html], but the gist of it is you will figure out which commits are yours (either by running `git log` on this branch, or `git log lidar_branch` if you're on another branch), make a new branch (perhaps something like `raad/lidar-decision-improved-twist-message-generation`), switch to it, and run `git cherry-pick HASH` for each commit thaat is yours, where HASH is the commit hash (that big string of numbers and letters in `git log`, or at least the first 7 characters of it), which will apply the changes made in that commit to your new branch. Sorry about the hassle, I really should have deleted this branch a while ago..... let me know when you're done so I can remove it from the git repo. At that point you'll be able to open a new PR on your new branch, and I'll be able to take a look at it.

raadkhan:
@garethellis0 I've gotten up to the step ```git cherry-pick HASH``` on a new branch, but I receive this:
```no changes added to commit
The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty
```
which makes it seem like the branch I just created already has all of the commits on the existing branch ```lidar_branch```

garethellis0:
@raadkhan - when you create a branch, you are branching off of you **current** branch. So if you want to create a fresh branch off of master (which is what you do here, you will have to `git checkout master`) and **then** `git branch ...`

PR #129, opened 2017-03-18
Improve obstacle merging

### LINKED ISSUE #107: Update lidar_decision to merge obstacles based on distance, as well as angular difference
- Currently, we decide to merge two lidar obstacles into the same one if the minimum difference in angle between them is below some threshold.
- We should also be considering how far away two objects are when we decide to merge them. The objects may appear to be right beside each other, but one may be 20m away, and another 3m away. We probably don't want to merge these two objects

lujinhao:
### Improving Obstacle Merging

- Wrote functions `getLastDistance` and `getFirstDistance`
- Add a parameter `max_obstacle_distance_diff` in `LindarDecision.h`
- Modified `lindar-decision-test.cpp` 

As a result, it can merge obstacles not only based on the angle but also based on the difference in distance between them.

closes #107


src/decision/include/LidarDecision.h
```diff
@@ -48,13 +48,27 @@ class LidarObstacle {
     LidarObstacle(std::vector<Reading> readings);
 
     /**
+    * Gets the distance of the first data from a Obstacle group
+    *
+    * @return the distance of the first data from a Obstacle group
+    */
+    distance_t getFirstDistance();
```
garethellis0: Since the vector of distances is sorted, this function is actually returning the **minimum** distance to the obstacle. As such, `getMinDistance` might be a better function name.

src/decision/include/LidarDecision.h
```diff
@@ -48,13 +48,27 @@ class LidarObstacle {
...
+
+    /**
+    * Gets the distance of the last data from a Obstacle group
+    *
+    * @return the distance of the last data from a Obstacle group
+    */
+    distance_t getLastDistance();
```
garethellis0: Again, the distances are sorted, so perhaps `getMaxDistance` might be a better function name.

src/decision/include/LidarDecision.h
```diff
@@ -140,7 +154,7 @@ class LidarDecision {
      * @return a vector of all obstacles
      */
     static std::vector<LidarObstacle> findObstacles(const sensor_msgs::LaserScan& scan,
-                                                        float max_obstacle_angle_diff);
+                                                        float max_obstacle_angle_diff, float max_obstacle_distance_diff);
```
garethellis0: just a formatting thing, each new variable should be on a new line, so this should be more like:
```
float max_obstacle_angle_diff, 
float max_obstacle_distance_diff);
```
rather then:
```
float max_obstacle_angle_diff, float max_obstacle_distance_diff);
```

src/decision/include/LidarDecision.h
```diff
@@ -159,8 +173,10 @@ class LidarDecision {
      * @param obstacles the list of obstacles to be merged (if similar)
      * @param max_angle_diff the max difference that two obstacles may differ by and still be considered
      *                      part of the same obstacle
+     * @param max_distance_diff the max difference that two obstacles may differ by and still be considered
+     *                      part of the same obstacle
      */
-    static void mergeSimilarObstacles(std::vector<LidarObstacle>& obstacles, float max_angle_diff);
+    static void mergeSimilarObstacles(std::vector<LidarObstacle>& obstacles, float max_angle_diff, float max_distance_diff);
```
garethellis0: Same as above, `max_distance_diff should be on a new line`

src/decision/include/LidarDecision.h
```diff
@@ -63,21 +77,21 @@ class LidarObstacle {
      *
      * @return the angle of the Obstacle from the robot
      */
-    float getAvgAngle();
+    angle_t getAvgAngle();
 
     /**
      * Gets the minimum angle from of an object from the robot
      *
      * @ return the minimum angle of the obstacle from the robot
```
garethellis0: Just realized that I messed up when I made this. While you're updating this, would you mind replacing `@ return` with `@return` throughout the file? Sorry about that! :disappointed:

src/decision/src/LidarDecision.cpp
```diff
@@ -29,6 +29,8 @@ LidarDecision::LidarDecision(int argc, char **argv, std::string node_name) {
     // Get Param(s)
     if (!public_nh.param<float>("max_obstacle_angle_diff", max_obstacle_angle_diff, (float)M_PI/36))
         ROS_WARN("max_obstacle_angle_diff not set, defaulting to %f", max_obstacle_angle_diff);
+    if (!public_nh.param<float>("max_obstacle_angle_diff", max_obstacle_angle_diff, (float)4))
+        ROS_WARN("max_obstacle_distance_diff not set, defaulting to %f", max_obstacle_angle_diff);
```
garethellis0: Probably just a copy-paste mistake, but I think you mean to try and get the "max_obstacle_distance_diff" param, rather then "max_obstacle_angle_diff" in this statement.

src/decision/src/LidarDecision.cpp
```diff
@@ -73,7 +75,7 @@ geometry_msgs::Twist LidarDecision::generate_twist_message(const sensor_msgs::La
 }
 
 std::vector<LidarObstacle> LidarDecision::findObstacles(const sensor_msgs::LaserScan &scan,
-                                                        float max_obstacle_angle_diff) {
+                                                        float max_obstacle_angle_diff, float max_obstacle_distance_diff) {
```
garethellis0: we probably want to have `float max_obstacle_distance_diff` on a new line here (same as previous comments)

src/decision/test/lidar-decision-test.cpp
```diff
@@ -7,6 +7,8 @@
 #include <LidarDecision.h>
 #include <gtest/gtest.h>
 
+#define distance 1.0f
```
garethellis0: generally speaking, in C++ code, we want to avoid defining constants like this. In this case, since you're only using it in `LidarObstacle` tests, I would suggest putting it in the `LidarObstacleTest` class declaration, **declaring** it where other variables like `scan1` are declared, and **defining** it in the `SetUp` function, similar to how `scan1` is declared.

src/decision/test/lidar-decision-test.cpp
```diff
@@ -116,27 +120,29 @@ class LidarDecisionTest : public testing::Test {
...
         EXPECT_GE(obstacle.getAvgDistance(), 20);
         EXPECT_LE(obstacle.getAvgDistance(), 40);
     }
 }
 
+
+
 int main(int argc, char **argv) {
     testing::InitGoogleTest(&argc, argv);
     ::testing::InitGoogleTest(&argc, argv);
```
garethellis0: I'd like to see at least one test where the obstacles are at a close enough `distance` to be merged, but **not** at a close enough `angle`, and one where the `angle` between them is small enough to be merged, but the `distance` is **not**. This ensures that both conditions for obstacle merging are being correctly checked for.

garethellis0: Note that you will like likely have to create new fake scans for these tests. You can declaring and fill them in the same way that `scan1` is declared and filled.

src/decision/src/LidarDecision.cpp
```diff
@@ -113,10 +117,11 @@ void LidarDecision::mergeSimilarObstacles(std::vector<LidarObstacle>& obstacles,
     int i = 0;
     while(i < (long)obstacles.size() - 1){
         // Check if angle difference between two consecutive scans is less then max_angle_diff
-        if (abs(obstacles[i+1].getMinAngle() - obstacles[i].getMaxAngle()) < max_angle_diff){
-            // Merge next obstacle into current one
-            obstacles[i].mergeInLidarObstacle(obstacles[i+1]);
-            obstacles.erase(obstacles.begin()+i+1);
+        if (abs(obstacles[i+1].getMinAngle() - obstacles[i].getMaxAngle()) < max_angle_diff && abs(obstacles[i+1].getFirstDistance() - obstacles[i].getLastDistance()) < max_distance_diff){
```
garethellis0: This should also be split across two lines, (split at the `&&`)

src/decision/src/LidarDecision.cpp
```diff
@@ -113,10 +117,11 @@ void LidarDecision::mergeSimilarObstacles(std::vector<LidarObstacle>& obstacles,
...
+                // Merge next obstacle into current one
+                obstacles[i].mergeInLidarObstacle(obstacles[i + 1]);
+                obstacles.erase(obstacles.begin() + i + 1);
+
```
garethellis0: Just a nitpicky thing, don't think we really need a space here.

src/decision/include/LidarDecision.h
lujinhao: Actually, I want to refer to the distance in pair of the minimum angle

src/decision/include/LidarDecision.h
lujinhao: Similarly, in pair of the max angle

src/decision/include/LidarDecision.h
lujinhao: I added a space for all `@` including `param`

src/decision/include/LidarDecision.h
lujinhao: changed

src/decision/include/LidarDecision.h
lujinhao: changed

src/decision/src/LidarDecision.cpp
lujinhao: yea, my mistake. It's changed now.

src/decision/src/LidarDecision.cpp
lujinhao: changed

src/decision/src/LidarDecision.cpp
lujinhao: Thank you. I will pay more attention to format in future

src/decision/src/LidarDecision.cpp
lujinhao: Spaces are deleted.

src/decision/test/lidar-decision-test.cpp
lujinhao: I simply used `1.0` directly similar to the angle parameters which are written as `0.02`.

src/decision/test/lidar-decision-test.cpp
lujinhao: I created scan 2 and a new unit test to test the case where distances are close enough but the angles are not. Similarly, scan 1 was changed to test the case where angles are close enough but the distances are not.

lujinhao:
@garethellis0 Thank you for your comments. I have made some changes accordingly.

src/decision/include/LidarDecision.h
garethellis0: Ah, ok, in that case, the `@return` value might be better phrased as: `the distance to the rightmost point of the obstacle`

src/decision/include/LidarDecision.h
garethellis0: same thing here, the `@return` value might be better phrased as: `the distance to the leftmost point of the obstacle`

src/decision/include/LidarDecision.h
garethellis0: Sorry, bad phrasing on my part I think, could you **remove** all spaces after `@`. ie. instead of `@ param` it should be `@param`

garethellis0:
@lujinhao - There were some merge conflicts that were quite complicated, so I've fixed them. Before you make any more changes, make sure that you pull my changes by running `git pull` when you are on the `improve-obstacle-merging` branch, to get my changes. After that, I just replied to one of your comments with a clarification. After you fix that, we should be good to merge. Sorry if that's a bit confusing, please message me via facebook messanger if you'd like some clarification or explanation.

Overall, excellent job, all changes were well made, code looks good :+1:

src/decision/include/LidarDecision.h
lujinhao: Fixed

src/decision/include/LidarDecision.h
lujinhao: Fixed

lujinhao:
@garethellis0 - I have changed the function documentation accordingly.

src/decision/include/LidarDecision.h
```diff
@@ -49,36 +49,50 @@ class LidarObstacle {
...
+    * Gets the distance to the rightmost point of the Obstacle
+    *
+    * @ return the distance to the rightmost point of the Obstacle
+    */
+    distance_t getFirstDistance();
+
+    /**
+    * Gets the distance to the leftmost point of the Obstacle
+    *
+    * @ return the distance to the leftmost point of the obstacle
```
garethellis0: Still need to remove the spaces between after `@` symbols. ie, `@return` instead of `@ return` and `@param` instead of `@ param`

garethellis0:
@lujinhao - can you address my final comment before saturday? I'd like to get this merged by then so we can get to testing this weekend.

lujinhao:
@garethellis0  Thank you for your feedback. I have removed all space after @.

PR #138, opened 2017-03-27
Added Params to the sample package 

### LINKED ISSUE #70: Add Example of getting params to sample_package
- an example of getting params can be found here: https://github.com/UBC-Snowbots/Snowflake/blob/master/src/decision/src/gps_decision/gps_decision.cpp
- Just a simple example should do, perhaps replacing the character that is added to the end of the string?

ValRat:
Closes  #70


ValRat:
ah i fugged up hold on.

PR #139, opened 2017-03-27
Added parameter to sample package and updated to latest Clion version

### LINKED ISSUE #70: Add Example of getting params to sample_package

ValRat:
Closes #70

### DIFF
--- a/src/sample_package/launch/rostest.launch
+++ b/src/sample_package/launch/rostest.launch
@@ -1,3 +1,0 @@
-<launch>
-    <node
-</launch>
--- a/src/sample_package/src/MyNode.cpp
+++ b/src/sample_package/src/MyNode.cpp
@@ -14,6 +14,11 @@ MyClass::MyClass(int argc, char **argv, std::string node_name) {
     ros::NodeHandle nh;
     ros::NodeHandle public_nh("~");
 
+    // Obtains character from the parameter server (or launch file), sets '!' as default
+    std::string parameter_name = "my_node/character";
+    std::string default_character = "!";
+    nh.param(parameter_name, suffix, default_character);
+
     // Setup Subscriber(s)
     std::string topic_to_subscribe_to = "subscribe_topic";
     int refresh_rate = 10;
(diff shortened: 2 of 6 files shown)


src/sample_package/launch/rostest.launch
```diff
@@ -1,3 +0,0 @@
-<launch>
-    <node
-</launch>
```
garethellis0: Why are we deleting this file? Are we getting rid of the rostest stuff?

src/sample_package/src/MyNode.cpp
```diff
@@ -14,6 +14,11 @@ MyClass::MyClass(int argc, char **argv, std::string node_name) {
     ros::NodeHandle nh;
     ros::NodeHandle public_nh("~");
 
+    // Obtains character from the parameter server (or launch file), sets '!' as default
+    std::string parameter_name = "my_node/character";
+    std::string default_character = "!";
+    nh.param(parameter_name, suffix, default_character);
```
garethellis0: We should be using `SB_getParam` here, from the kitchen sink library :wrench: . I've already added all the stuff to the cmake, should just be a matter of using it. Example [here](https://github.com/UBC-Snowbots/IGVC-2017/blob/master/src/decision/src/LidarDecision.cpp)

src/sample_package/launch/rostest.launch
ValRat: This was mistakenly committed in, rostests are defined under test/sample_package_test.test

src/sample_package/src/MyNode.cpp
ValRat: Fixed :hammer_and_wrench:

src/sample_package/launch/rostest.launch
ValRat: This was mistakenly committed in, rostests are defined under test/sample_package_test.test

PR #143, opened 2017-04-01
Added an imu to the jfrost model

garethellis0:
Wow. Git really mangled this one.... please check that gazebo still works, but all I did is add the new imu section, and fix the tabbing on the section above.


garethellis0:
When you're testing this, please make sure that you get sane readings from the imu, especially the compass, that match this coordinate system:
```
              +X
              ^
              |
      +θ  +<----->+ -θ
          |   |   |
          V   |   V
+Y <---------------------> -Y
              |
              |
              |
              V
              -X
```

garethellis0:
@ValRat - I had to merge gps driver stuff into this branch, so it is stalled on #122 - please don't merge this guy until we merge #122 (when GPS stops being stupid at the EDC)

install_dependencies.sh
```diff
@@ -24,8 +23,11 @@ sudo apt-get install -y\
     ros-kinetic-effort-controllers \
     ros-kinetic-hector-gazebo \
     ros-kinetic-turtlebot-teleop \
+    python-rosinstall \
+    python-catkin-tools \
```
ValRat: might wanna take this out before final merge

garethellis0:
@ValRat - whoops. that would be the other compass plugin I was previously using..... an IMU message only publishes the heading as a quaternion: http://docs.ros.org/api/sensor_msgs/html/msg/Imu.html, and does not have a `compass` subtopic

garethellis0:
@ValRat - would you mind testing out the imu over on `gareth/gps_decision`? It should be publishing to the `/imu` topic... if it works there, then fine, the IMU can be added via that PR.

garethellis0:
apologies... to manually test it, you'll need to convert the quaternion to degrees. this may help: http://www.onlineconversion.com/quaternions.htm

PR #144, opened 2017-04-01
Fixed coordinate system in gpsmanager

garethellis0:
Please run this through gazebo once to confirm to yourself it's not broken (will depend on #143)


garethellis0:
le sigh.... ignore this for now. I need to confirm if present me or past me was right.... :hourglass_flowing_sand:

ValRat:
The greatest enemy is yourself

PR #151, opened 2017-04-12
Removed the src/firmware folder

garethellis0:
We don't need the `src/firmware` folder, since we've moved to our own fork.


ValRat:
:hammer:

PR #162, opened 2017-04-20
Fixed Lat and Lon Values in gazebo being switched

garethellis0:
I think I'm sane now, not sure what I was smoking when I set this up, but everything should be good now.
To confirm that I am not on a massive trip, please:

- ensure that the compass heading in gazebo (the heading from the IMU, it's a quaternion, sorry.... maybe just break in the `imuCallback` in `GpsManager.cpp`? It auto converts it to heading there....) matches up with the lon and lat values. 
image


ValRat:
Wait I thought the compass output was sane with the last pull request...

garethellis0:
Yup, I believe that the compass output is totally fine right now, I just want to make sure that I'm sane when I see that it lines up with the lat/lon. 

So, presuming that an IMU heading of 0 is pointing directly north, and we are at lat/lon coordinates in the upper western hemisphere, make sure that moving the robot around changes the lat/lon as you would expect. **ie.** moving north should increase latitude, moving east should increase longitude (though it's negative, so increasing towards 0)

src/sb_gazebo/urdf/jfrost.gazebo
```diff
@@ -162,8 +162,8 @@
             <!-- Set these parameters depending on where you want to gps reference to be -->
             <!-- NOTE: THESE SHOULD MATCH THE COMPASS_SIM PLUGIN - SEE BELOW-->
             <!-- EDC (Very Roughly) -->
-            <referenceLongitude>49.262479</referenceLongitude>
```
ValRat: `<!-- NOTE: THESE SHOULD MATCH THE COMPASS_SIM PLUGIN - SEE BELOW-->`
If you're changing these up change the ones below as well?

ValRat: Fixed

garethellis0:
@ValRat - was the conclusion on this that the initial changes were correct?

garethellis0:
@ValRat - was the conclusion on this that the initial changes were correct?

ValRat:
Hi yes, took a nap, meditated on it, did a final sanity check and what you had before was correct. Removed some of the more misleading topics and files.

src/sb_gazebo/urdf/jfrost.gazebo
ValRat: Fixed the fix

Issue #170, opened 2017-04-30
Create a node which transforms Gazebo's Zed point cloud output internally


ValRat:
Currently, we specify a transform in `jfrost.xacro` in order to re-orient the output of the "Zed" module of the jfrost simulation to the proper place in the world. But this means that intermediate nodes which accesses this data *cough* `zed_filter` *cough* uses the old pre-transformed point-cloud data which results in wonky and incorrect data output.

We should consider writing a node which transforms the Zed's point cloud output internally so that all data which depends on it are using correct world-frame coordinates. See `jfrost.xacro` for the specific transform hack.

garethellis0:
closed by #175

PR #175, opened 2017-05-03
Reoriented the Gazebo Zed's pointcloud

ValRat:
Note of 
```
tfBuffer.lookupTransform(msg->header.frame_id, "zed_pointcloud", ros::Time(0), ros::Duration(1.0));
```
where the `target_frame` and `source_frame` seems to be switched. I swear to god the transform is proper. So either this is a bug, can't confirm since my tf knowledge is abysmal, or I'm dumb and I'm not declaring frames correctly (more likely).
Anyhow it all works :tm: .

### DIFF
--- a/src/sb_gazebo/CMakeLists.txt
+++ b/src/sb_gazebo/CMakeLists.txt
@@ -6,6 +6,8 @@ project(sb_gazebo)
 ## is used, also find other catkin packages
 find_package(catkin REQUIRED COMPONENTS
   gazebo_ros
+  tf2_ros
+  tf2_sensor_msgs
 )
 
 add_definitions(-std=c++14)
@@ -27,6 +29,7 @@ include_directories(
 
 ## Declare a C++ executable
 # add_executable(gazebo_node src/gazebo_node.cpp)
+add_executable(zed_transform src/zed_transform.cpp)
 
 ## Specify libraries to link a library or executable target against
 # target_link_libraries(gazebo_node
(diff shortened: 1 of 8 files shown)


garethellis0:
We should talk about this tomorrow, I think I understand what's going on, but tf's are breaking my brain. Just going to leave some notes here, @ValRat, please let me know if these are correct:

- For some reason, the pointcloud published by the gazebo sim node is not in the frame of the camera (or at least it's rotated 90 degrees on multiple axis). Ideal solution here is to change the plugin itself so that we're actually publishing the pointcloud to the correct frame. For the IMU plugin we have these options, any chance there are similar ones for the zed plugin?:
```
      <xyzOffsets>0 0 0</xyzOffsets>
      <rpyOffsets>0 0 0</rpyOffsets>
```
- Secondary solution is to re-publish the pointcloud in the correct orientation, by publishing a static transform (of just some rpy values), and then applying that to the pointcloud, which we then re-publish. In this case, we should have the plugin publish to some garbage value (say `/zed/camera/point_cloud/uncorrected_pointcloud_do_not_use` or something) and the pointcloud transformed to the camera's tf should be on the topic the zed would normally publish to (`/camera/point_cloud/cloud_registered `). Incidentally, everything from the zed should be under `/camera` not `/zed` (at least according to [the doc](http://wiki.ros.org/zed-ros-wrapper))

ValRat:
Gucci 👌

garethellis0:
closes #170

PR #179, opened 2017-05-13
Setup an Extended Kalman Filter with robot_pose_ekf

garethellis0:
**DO NOT MERGE** Just being used to track progress

Created a launch file to start up nodes to transform NavSatfix to Odom and run `robot_pose_ekf`


garethellis0:
@ValRat - once gazebo is up and running, all the topics should already be in place to get this to work. Can you confirm that, after running gazebo and `localisation/launch/ekf.launch`, you can see the `odom_combined<->base_link` transform in the tf tree and in rviz. I think if that works for you, we should be basically good to go for testing it out thursday (although we may have more sensor-specific issues, will try and test tomorrow).

garethellis0:
The equivalent changes to this PR ended up making their way to master through #192

PR #181, opened 2017-05-14
Update gps_manager initialization

garethellis0:
- `gps_manager` will now take the average of the first `n` readings it receives (where `n` is seperate params for heading and navsatfix) and average them to come up with our initial heading and navsatfix

### DIFF
--- a/src/decision/include/GpsManager.h
+++ b/src/decision/include/GpsManager.h
@@ -29,7 +29,7 @@ struct Waypoint {
 
 class GpsManager {
 public:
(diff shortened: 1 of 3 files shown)


garethellis0:
@garethellis0 - test this out with gps

garethellis0:
@ValRat - we should be good to go (pending code review). I tested this out in Gazebo and all seemed well. Robot still performed pretty well.

garethellis0:
We ended up using a much better method that constantly recalculated everything, rather then relying on getting a good reading from the start. As such, I'm closing this one.

PR #187, opened 2017-05-25
created final_decision launch file

garethellis0:


garethellis0:
Competition is now over, and the equivalent to this launch file has made it into master. Closing.

PR #190, opened 2017-08-14
Competition

garethellis0:
**DO NOT MERGE, JUST CREATING SO WE CAN SEE WHAT WE'RE DOING**


garethellis0:
was only ever intended to provide nice access to github diff's, so closing

PR #191, opened 2017-08-14
Added the `gps_common` package

garethellis0:
See `README.md` for details.


ValRat:
Remind me why we're not submoduling their repo/using the official ros package?

garethellis0:
I believe the original reason was that we switched east and north vs. x and y, but that's honestly a terrible idea, seeing as `gps_common` uses coordinate frames properly, as far as I can tell. I'll fix the issue properly elsewhere. Good catch.

PR #195, opened 2017-08-19
New `gps_manager` and `gps_decision` nodes, and some general cleanup

### LINKED ISSUE #94: Move away from using install_dependencies to rosdep
Right now we're using a script (`install_dependencies`)to manually install all dependencies, but CI uses a different (and much better system) where it finds all the requirements from the `package.xml` files (fairly sure it's rosdep, or something like that anyhow). We should look at what CI is doing (script is located [here](https://github.com/UBC-Snowbots/moveit_ci)) and change `install_dependencies` to just use the same commands that CI does (although it should be kept separate from `getting_started` to make it easier for people to update all their dependencies **after** they've setup the repo for the first time

### LINKED ISSUE #185: Phase out gps_decision and gps_manager
We've created updated nodes that use TF's properly, namely `GpsManager2` and `GpsDecision2`. We need to:
- removed the old nodes, rename the new nodes to the old names,
- update the launch files
- update the cmakelist
- check that it compiles
- test them out in the sim

This all depends on the new nodes getting in, but they seem better, so this issue is very likely valid

garethellis0:
This is a pretty big one, so I've added some comments to files to help you as you read through, but generally, here's what's been changed:

- Totally re-wrote `gps_manager` and `gps_decision`. You probably want to just review the files associated with these two nodes as if I was adding them for the first time, because the diffs  are absolutely **gross** (resolves #185)
- Fixed the setup script to use rosdeps (resolves #94)

(Also *wow*, I really should have made that a review instead of just individual comments. Sorry about that)


.gitignore
```diff
@@ -18,3 +18,6 @@ cmake-build-debug
 *_REMOTE_*.txt
 *.orig
 
+# Symbols database for stuff like autocomplete in Vim
+compile_commands.json
+
```
garethellis0: This file gives me autocomplete in Vim!

.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule "src/nmea_navsat_driver"]
 	path = src/nmea_navsat_driver
 	url = https://github.com/UBC-Snowbots/nmea_navsat_driver.git
+[submodule "src/gps_umd"]
+	path = src/gps_umd
+	url = https://github.com/swri-robotics/gps_umd
```
garethellis0: The `gps_umd` repo contains the `gps_common` package, which is being used by `gps_manager` to convert lat/lon into UTM coordinates (x and y in meters)

install_dependencies.sh
```diff
@@ -14,19 +14,18 @@ echo "================================================================"
...
+
+# The current directory
+CURR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Install all required dependencies to build this repo
+rosdep install --from-paths src --ignore-src --rosdistro kinetic -y
+
+# We have this extra rosdep call because rosdep can't seem to find nested
+# `package.xml` files, like those in src/gps_umd/gps_common
+rosdep install --from-paths src/gps_umd --rosdistro kinetic -y
```
garethellis0: This switches us over from using a list of packages to install to using rosdep's, which basically check the dependencies listed in each `package.xml` file and installs them. It has some issues if they're deeply nested (as is the case with those in `gps_umd`) but otherwise this is a much better system that should make sure we can keep dependencies updated without having to worry about maintaining a list. This should also resolve #94!

src/decision/CMakeLists.txt
```diff
@@ -36,11 +37,35 @@ include_directories(
...
+add_executable(final_decision 
+    src/final_decision.cpp
+    src/FinalDecision.cpp 
+    include/FinalDecision.h
+    )
+add_executable(gps_manager 
+    src/gps_manager.cpp
+    src/GpsManager.cpp 
+    include/GpsManager.h
+    )
```
garethellis0: I'm thinking about setting this as standard formatting for all `CMakeLists.txt` files, but that's pretty hard to enforce, so not sure if it's feasible :cry:

src/decision/include/GpsDecision.h
```diff
@@ -1,9 +1,9 @@
 /*
  * Created By: Gareth Ellis
- * Created On: April 1, 2017
+ * Created On: May 18, 2017
```
garethellis0: Heh.

src/decision/include/GpsManager.h
```diff
@@ -1,19 +1,20 @@
 /*
  * Created By: Gareth Ellis
- * Created On: October 20, 2016
+ * Created On: May 17, 2017
```
garethellis0: Heh.

src/decision/include/GpsManager.h
```diff
@@ -1,19 +1,20 @@
...
  * Description: Manages the GPS waypoints we are given
  *              (publishing the next one once we've arrived
  *              at the current one)
  */
```
garethellis0: Wow. The diff on this file looks *TERRIBLE*. What happened here is I basically re-wrote it from scratch, so I suggest you just review the file by itself as if I were submitting it for the first time.

src/decision/src/GpsDecision.cpp
```diff
@@ -1,9 +1,9 @@
 /*
- * Created By: Gareth Ellis and Chris Chen
- * Created On: September 22, 2016
+ * Created By: Gareth Ellis
+ * Created On: May 18, 2017
```
garethellis0: HAHAHAHAHAH NOT AT ALL CLOSE TO COMPETITION HAHAHAHA :sob:

src/decision/src/GpsDecision.cpp
```diff
@@ -1,9 +1,9 @@
...
  * Description: The Decision Node for GPS, takes in a point relative to
  *              the robots location and heading and broadcasts a
- *              recommended twist message
+ *              recommended twist message. Revised to use TF's
  */
```
garethellis0: Same thing with this file as with `GPSManager.cpp`, probably want to just review it as if I were submitting it for the first time

src/decision/src/GpsManager.cpp
```diff
@@ -1,29 +1,11 @@
 /*
  * Created By: Gareth Ellis
- * Created On: October 20, 2016
+ * Created On: May 17, 2017
```
garethellis0: hehe

src/decision/src/GpsManager.cpp
```diff
@@ -1,29 +1,11 @@
...
  * Description: Manages the GPS waypoints we are given
  *              (publishing the next one once we've arrived
  *              at the current one)
  */
```
garethellis0: Same thing here, basically new file, probably want to just review as a new submission, rather then via the diff.

src/decision/src/gps_decision.cpp
```diff
@@ -1,9 +1,9 @@
 /*
- * Created By: YOUR NAME HERE
- * Created On: September 22, 2016
+ * Created By: Gareth Ellis
+ * Created On: May 18, 2017
```
garethellis0: weeeeeeee

src/decision/test/gps-manager-test.cpp
```diff
@@ -49,206 +49,6 @@ TEST(GpsManager, parseWaypoints){
...
-    nsfix.longitude = -123.240825;
-    wp.lat = 49.25431591;
-    wp.lon = -123.240825;
-    double origin_heading = 60 * (M_PI/180);
-
-    geometry_msgs::Point converted_waypoint = GpsManager::convertToRobotsPerspective(wp, nsfix, origin_heading);
-    EXPECT_NEAR(10 * cos(origin_heading), converted_waypoint.x, 0.005);
-    EXPECT_NEAR(-10 * cos(M_PI/2 - origin_heading), converted_waypoint.y, 0.005);
-}
-
```
garethellis0: It's sad that these tests are vanishing, because tests make me happy, but on the other hand, we're removing them because we're just relying on `TF`'s to handle all this properly for us now, so I'm OK with that...

src/sb_gazebo/launch/gazebo_open_world.launch
```diff
@@ -0,0 +1,57 @@
+<launch>
+  <!--
+    Launch File For Opening Worlds - From the ros_gazebo package
+    
+    Snowbots Modifications:
+        Added optional `tf` argument so we can remap the tf topic for gazebo to prevent tf tree conflicts
```
garethellis0: This is the main reason I added this. We wanted to be able to generate and use the TF tree based solely off our localisation system, but a bunch of the plugins that we use in Gazebo publish TF's that interefere, so we solve that problem by giving them their own TF tree.

ValRat:
^ Woah now XD

.gitignore
garethellis0: (but we're ignoring it because it's generated by the compiler at build)

src/decision/test/gps-manager-test.cpp
ValRat: Trust in the libs :book:

src/sb_gazebo/urdf/jfrost.gazebo
```diff
@@ -151,13 +151,12 @@
...
       <velocityDrift>0 0 0</velocityDrift>
       <velocityGaussianNoise>0.1 0.1 0.1</velocityGaussianNoise>
-      <!-- Set these parameters depending on where you want to gps reference to be -->
-      <!-- NOTE: THESE SHOULD MATCH THE COMPASS_SIM PLUGIN - SEE BELOW-->
       <!-- EDC (Very Roughly) -->
       <referenceLatitude>49.262479</referenceLatitude>
       <referenceLongitude>-123.248600</referenceLongitude>
       <referenceAltitude>50.00</referenceAltitude>
-      <referenceHeading>0.00</referenceHeading>
+      <!-- referenceHeading is in DEGREES -->
```
ValRat: Heh

src/decision/src/FinalDecision.cpp
```diff
@@ -24,8 +24,8 @@ FinalDecision::FinalDecision(int argc, char **argv, std::string node_name) {
     gps_subscriber = public_nh.subscribe(gps_decision_topic_name, queue_size, &FinalDecision::gpsCallBack, this);
 
     // Setup Publisher(s)
-    std::string twist_topic = public_nh.resolveName("twist");
-    twist_publisher = nh.advertise<geometry_msgs::Twist>(twist_topic, queue_size);
+    std::string twist_topic = public_nh.resolveName("cmd_vel");
```
ValRat: Is `resolveName` needed with a public nodehandle? I think it just brings it to the global namespace

src/decision/src/GpsDecision.cpp
ValRat: XD

ValRat:
I've written what jumped out at me when I stared at the code, will do functional test tomorrow (gotta get my desktop ubuntu updated).

src/decision/src/GpsDecision.cpp
```diff
@@ -17,12 +17,6 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
 
     // Setup Subscriber(s)
     uint32_t refresh_rate = 10;
```
garethellis0: Right on both counts, changed!

src/decision/src/GpsDecision.cpp
```diff
@@ -17,12 +17,6 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
-    std::string current_location_topic = "/gps_manager/current_location";
-    current_location_subscriber = private_nh.subscribe(current_location_topic, refresh_rate,
-                                                    &GpsDecision::currentLocationCallback, this);
-    std::string imu_topic = "/gps_manager/current_heading";
-    imu_subscriber = private_nh.subscribe(imu_topic, refresh_rate,
-                                          &GpsDecision::imuCallback, this);
     std::string waypoint_topic = "/gps_manager/current_waypoint";
     waypoint_subscriber = private_nh.subscribe(waypoint_topic, refresh_rate,
```
garethellis0: Good point, changed to `nh`

src/decision/src/GpsDecision.cpp
```diff
@@ -36,29 +30,48 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
+    SB_getParam(private_nh, "global_frame", global_frame, (std::string)"odom_combined");
 }
 
-void GpsDecision::waypointCallback(const geometry_msgs::Point::ConstPtr &waypoint) {
-    // Create a new twist message and publish it
-    geometry_msgs::Twist twist = mover.createTwistMessage(current_location, current_heading, *waypoint);
-    twist_publisher.publish(twist);
+void GpsDecision::waypointCallback(const geometry_msgs::PointStamped::ConstPtr &waypoint) {
+    // Get the current position and heading of the robot
+    tf2_ros::Buffer tfBuffer;
```
garethellis0: Made into class members.

src/decision/src/GpsDecision.cpp
```diff
@@ -36,29 +30,48 @@ GpsDecision::GpsDecision(int argc, char **argv, std::string node_name) {
...
+    tf2_ros::TransformListener tfListener(tfBuffer);
+    try {
+        geometry_msgs::TransformStamped tfStamped = tfBuffer.lookupTransform(
+                global_frame, base_frame, ros::Time(0), ros::Duration(1.0));
```
garethellis0: Weeeeellll, that would imply `waypoint.header.stamp` was set correctly.... which it wasn't...

In any case, it's set correctly now, and we're using it as suggested :100:

src/decision/src/GpsManager.cpp
```diff
@@ -33,134 +15,63 @@ GpsManager::GpsManager(int argc, char **argv, std::string node_name){
...
-    uint32_t refresh_rate = 1;
-    std::string raw_gps_topic_name = "/gps_driver/navsatfix";
-    raw_gps_subscriber = private_nh.subscribe(raw_gps_topic_name, refresh_rate,
-                                             &GpsManager::rawGpsCallBack, this);
-    std::string imu_topic_name = "/imu";
-    imu_subscriber = private_nh.subscribe(imu_topic_name, refresh_rate,
-                                             &GpsManager::imuCallback, this);
+    uint32_t queue_size = 1;
+    std::string tf_topic_name = "/tf";
+    tf_subscriber = private_nh.subscribe(tf_topic_name, queue_size,
```
garethellis0: changed to `nh`

src/decision/src/FinalDecision.cpp
garethellis0: Good point, changed.

src/decision/src/GpsDecision.cpp
garethellis0: Gahh, tried this and remembered why I didn't do it in the first place. We can only create the `TransformListener` **after** we run `ros::init`. I could resolve this by making the class member a pointer, which I could initialise in the constructor, but given that this isn't that performance sensitive, and if we're not getting a brand new TF we shouldn't be moving anyhow, I'm going to leave this as is.

src/decision/src/GpsDecision.cpp
garethellis0: okie, so now I remember why this doesn't work either... since we can't make the `TransformListener` a class member, we publish the waypoint and set the stamp to the current time, but we only start listening for TF's after we get the waypoint, and we can't get TF's from the past, so we crash. On the other hand, when deciding on how to move, we want to use the most up to date TF to determine the current state of the robot anyways, so using `ros::Time(0)` is probably the right choice.

garethellis0:
OK, changes made and tested out in Gazebo, everything still seems to work fine, ready for re-review!

ValRat:
Had to do a clean install of the Repo to deal with submodule shenanigans but node works in sim. PR is merge-able.

garethellis0:
Nice work here @ValRat! Looks like everything is good to go?

PR #196, opened 2017-08-23
Added Enforced Formatting to CI via clang-format

### LINKED ISSUE #167: Enforce Code Formatting
Enforcing code formatting, like using 4 spaces or brackets on the right lines, would probably go a long way towards keeping the codebase readable. We should look at using something like [clang-tidy](http://clang.llvm.org/extra/clang-tidy/) or [clang-format](https://clang.llvm.org/docs/ClangFormat.html), for this (we would need a separate tool for our XML). We hopefully would be able to make it check code as part of CI, and make an easy command that people can use to fix formatting in their code if it fails CI.

garethellis0:
@ValRat - this should be good to go:
- updated the install script to actually use rosdeps
- added some utilities to make working with clang format easier (`check_formatting.sh` and `fix_formatting.sh`)
- added binaries for `clang-format` and `git-clang-format` so that we're always using the same version
- updated `README.md` with some info on how to use the new utilities
- updated CI to actually check for clang-format as well
- resolves #167 

Currently, we're only checking changed files for clang-format.

For testing, maybe just branch off of this branch, make a change that will cause clang-format to fail, go through the errors on Travis, make sure they make sense, 
and then check and fix it with the new utilities.


ValRat:
Does this close the CI dependency issue that I was supposed to be working on?

garethellis0:
Sadly no :(

ValRat:
Fixed and tested :+1:

PR #221, opened 2017-10-03
Updated and Simplified the clang-format scripts

garethellis0:
- removed \`check_formatting.sh\`, as you can just run \`fix_formatting.sh\` to check and fix anything that's needed (it will refuse to overwrite uncommitted changes)
- update \`fix_formatting.sh\` to take the branch to compare against (to determine what changes to run \`clang-format\` over). This was needed now that we're no longer just merging into one branch


ValRat:
Looks good! Thanks for :wrench:

Issue #113, opened 2017-03-03
Rename ros_vision 


garethellis0:
- currently our IPM (inverse prespective mapping) and HSV filters are in a node/class called `ros_vision`
- this is not a very good name. Yes we use ros, and yes it does have to do with vision, but we should probably rename it to something more reflective of what it actually does
- a suggestion could be `ipm_and_hsv_filter`
- this will likely require a change in at minimum: the `CMakeLists.txt`, the class name in `.cpp` and `.h` files for  ros_vision, the names of aforementioned `.cpp` and `.h` files, the topics that other nodes are subscribing to that are under the ros vision namespace (ie. `/ros_vision/something`), launch files...

garethellis0:
Resolved in #223

PR #223, opened 2017-10-03
Migrated Vision Upgrades from IARRC-2017 to Snowflake

garethellis0:
- Removed old `ros_vision` stuff
- Added new `HSV` and `IPM` nodes
- merged code was from `IARRC 2017` - `robyn/working_merged_code`


src/vision/src/HSVFilterNode.cpp
```diff
@@ -0,0 +1,177 @@
+/*
+ * Created By: Robyn Castro
+ * Created On: June 17, 2017
+ * Description: Filters out everything but green.
```
ValRat: More apt description would be like:
`Filters an image in the HSV colourspace to a binary image`

src/vision/src/ipm_filter.cpp
```diff
@@ -0,0 +1,20 @@
...
+ * Description: Filters out all non-green from the image
+ *              then publishes the new filtered image.
+ */
+
+#include "../include/IPMFilterNode.h"
+#include <ros/ros.h>
+
+int main(int argc, char** argv) {
+    // Setup your ROS node
+    std::string node_name = "IPMFilter";
```
RobynCastro: Just realized hsv_filter node_name is in snake case. For consistency sake, should IPMFilter be changed or hsv_filter?

RobynCastro:
Looks good, all files seem to be there, and all changes seem reasonable. Just have one minor comment on node_names inside hsv_filter.cpp and ipm_filter.cpp

ValRat:
This closes #113 as well.

src/vision/src/ipm_filter.cpp
garethellis0: Good point. I'm not sure if we've every explicitly looked at this and decided before, but most nodes, including the sample node, seem to be in snake case. I've changed this here, as well as in `vision/CMakeLists.txt`.

garethellis0:
@RobynCastro - ready for re-review

PR #238, opened 2017-10-11
Merging core into iarrc

garethellis0:


garethellis0:
Looks like we can't resolve conflicts here. Will have to open my own branch

PR #240, opened 2017-10-11
Merging core into igvc

garethellis0:


garethellis0:
damit. wrong direction

PR #241, opened 2017-10-11
Merging core into igvc

garethellis0:


garethellis0:
The review process here should be a fairly simple matter of making sure that all the changes being merged over are sane (ie. we want them in `igvc`)

.gitignore
```diff
@@ -18,6 +18,3 @@ cmake-build-debug
 *_REMOTE_*.txt
 *.orig
 
-# Symbols database for stuff like autocomplete in Vim
```
ValRat: I'm gonna guess this is intentional?

garethellis0: Yeah, we now have a symlink there for the `compile_commands.json` file, so we don't want to ignore it anymore.

garethellis0:
@ValRat @raadkhan - ready for re-review

PR #253, opened 2017-11-18
Rviz Utilities

### LINKED ISSUE #252: Create an Rviz Utilities Class
Takes in objects such as points, and creates an Rviz marker of it.

RobynCastro:
Resolves #252


src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+#ifndef SB_UTILS_RVIZUTILS_H
+#define SB_UTILS_RVIZUTILS_H
+
+// Messages
+#include <geometry_msgs/Point.h>
+#include <visualization_msgs/Marker.h>
+
+using namespace std;
+using namespace visualization_msgs;
```
garethellis0: `using namespace` lines should be in the `.cpp` file. This does mean the signatures and fields in the header will need to be changed, but If they're in the header, then anything that includes the header will automatically get them, which can cause conflicts between things with the same name (say if we defined a class called `Marker`, then it would conflict with `visualization_msgs::Marker` if the file using `Marker` included this header)

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+
+class RvizUtils {
+public:
+    /**
+     * Turn the points into a marker for rviz
+     *
+     * @param points the points to be converted
+     * @param color the color of the points
```
garethellis0: Missing a couple of `@param`'s here

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+     */
+    static Marker displayPoints(vector<geometry_msgs::Point> points, Marker::_color_type color,
+                                Marker::_scale_type scale, string frame_id, string ns);
+
+    /**
+     * Turn the point into a marker for rviz
```
garethellis0: nit-pick, but this should say `a point`, rather then `the point`

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param point the points to be converted
```
garethellis0: should be "the **point** to be converted*

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+     * @param color the color of the point
```
garethellis0: Missing some `@param`'s and a `@return`

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+    static Marker displayPoint(geometry_msgs::Point point, Marker::_color_type color, Marker::_scale_type scale,
+                               string frame_id, string ns);
+
+    /**
+     *  Create a marker color type based on given red, green, blue, alpha values.
+     *
+     *  @param r red
+     *  @param g green
+     *  @param b blue
+     *  @param a alpha
```
garethellis0: Missing an `@return`

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+     */
+    static Marker::_color_type createMarkerColor(float r, float g, float b, float a);
+
+    /**
+     *  Create a marker scale type based on given x, y, and z scales.
+     *
```
garethellis0: Missing `@param`'s

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,65 @@
...
+     */
+    static Marker::_scale_type createrMarkerScale(float x, float y, float z);
+private:
+
```
garethellis0: We should make the constructor private, as there is no reason for anyone to construct an instance of this class. You can do that by just adding `RvizUtils() = 0` here I think.

RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
RobynCastro: Fixed

src/sb_utils/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ include_directories(
 ## Declare a C++ library
 add_library(sb_utils
     include/sb_utils.h src/sb_utils.cpp
+    include/RvizUtils.h src/RvizUtils.cpp
```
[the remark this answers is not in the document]
RobynCastro: Fixed

src/sb_utils/include/RvizUtils.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ */
+#ifndef SB_UTILS_RVIZUTILS_H
+#define SB_UTILS_RVIZUTILS_H
+
+// Messages
+#include <geometry_msgs/Point.h>
+#include <visualization_msgs/Marker.h>
+
+class RvizUtils {
```
garethellis0: OK, so after some discussion in the slack #software-leads channel, we've decided to place `RvizUtils` under the `snowbots` namespace. After that, this should be good to go.

PR #224, opened 2017-10-03
Moved Green Light/Circle Detection from IARRC 2017 to Snowflake

garethellis0:
- **Note:** the tests here depend on `HSVFilter`, so this won't pass CI until we merge #223, then merge `core` into `iarrc`

PR #259, opened 2017-12-19
Merging iarrc and core into igvc

garethellis0:
After some discussion, we've decided to go back to a single branch repo structure. The process for this migration is the following

1) Merge relevant changes from `iarrc` and `core` into `igvc`
2) Rename folders to make it clear what belongs to `igvc`, `iarrc` or neither
3) Rename `igvc` to `master`

This PR is for (1) and (2). We can do (3) once this is merged, but this way we can at least see a nice diff of what's being changed, which hopefully should let us catch any nasty bugs induced through this process.


src/drag_race_iarrc/include/LidarObstacle.h
```diff
@@ -0,0 +1,245 @@
+/*
+ * Created By: Gareth Ellis
+ * Created On: January 22, 2017
+ * Description: This class represents an Obstacle, as detected by the lidar,
+ *              and provides various utilities for working with obstacles
+ */
+
+#ifndef LIDAROBSTACLE_H
```
RobynCastro: Convention for this line in each of the files is all over the place. Some have the package name included and some don't. Not really sure what the proper convention is though haha, but if the package name is included, it should now be DRAG_RACE_IARRC_xxxxx_H.

src/vision/test/ros-vision-test.cpp
```diff
@@ -1,13 +0,0 @@
-//
-// Created by valerian on 15/10/16.
```
raadkhan: we should formalize this

src/sb_vision/src/ipm_filter.cpp
```diff
@@ -0,0 +1,20 @@
+/*
+ * Created By: Robyn Castro
+ * Created On: June 15, 2017
+ * Description: Filters out all non-green from the image
+ *              then publishes the new filtered image.
+ */
+
+#include "../include/IPMFilterNode.h"
```
raadkhan: nitpicky but we can avoid bugs

src/sb_vision/src/hsv_filter.cpp
```diff
@@ -0,0 +1,20 @@
+/*
+ * Created By: Robyn Castro
+ * Created On: June 15, 2017
+ * Description: Filters out all non-green from the image
+ *              then publishes the new filtered image.
+ */
+
+#include "../include/HSVFilterNode.h"
```
raadkhan: here too

src/sb_vision/src/HSVFilterNode.cpp
```diff
@@ -0,0 +1,177 @@
+/*
+ * Created By: Robyn Castro
+ * Created On: June 17, 2017
+ * Description: Filters an image in the HSV colourspace to a binary image
+ *
+ */
+
+#include "../include/HSVFilterNode.h"
```
raadkhan: here as well

src/sb_vision/src/HSVFilter.cpp
```diff
@@ -0,0 +1,101 @@
...
+ * Takes in an image and transforms it into a binary image
+ * given some color specification.
+ * Author: Valerian Ratu
+ * Ref: 	Color Picker
+ *			https://raw.githubusercontent.com/kylehounslow/opencv-tuts/master/auto-colour-filter/AutoColourFilter.cpp
+ *		Color Bar:
+ *			http://opencv-srf.blogspot.ca/2010/09/object-detection-using-color-seperation.html
+ */
+
+#include "../include/HSVFilter.h"
```
raadkhan: not sure if the convention is `include_library( )` or `include_directories( )` to include headers in the `CMakeLists.txt` but I think will be better than using relative path for headers

ValRat:
Sorry for the very tardy review, everything seems OK and working on my first pass.
Should probably do a full-repo clang format at some point - otherwise all OK on my end. We're not deleting either branches just yet so if anything seems amiss can merge them in afterwards.

Thanks for doing this! 

Another thing to probably watch out for is that #224 would need to be re-done/re-targeted.

src/drag_race_iarrc/include/LidarObstacle.h
garethellis0: Hmm, that's a good point, doesn't look like we've done a very good job at keeping this consistent. :sweat_smile: In this case I think your idea here seems pretty good. It matches up with what we're doing in the sample package, and is unique enough that we are unlikely to have conflicts (with multiple headers having the same header guards).

src/sb_vision/src/HSVFilter.cpp
garethellis0: Good catch! We should definitly be using relative includes here (which works because of the exact reason you mentioned). :+1: Changed

src/sb_vision/src/HSVFilterNode.cpp
garethellis0: changed

src/sb_vision/src/hsv_filter.cpp
garethellis0: changed

src/sb_vision/src/ipm_filter.cpp
garethellis0: Not nitpicky at all! This is exactly the sort of thing these reviews are supposed to catch.

src/vision/test/ros-vision-test.cpp
garethellis0: I think we already have, is what you mean at the start of "Coding Conventions" in `README.md`?

garethellis0:
@RobynCastro @ValRat @raadkhan - requested changes made, just need to get a approval again when one of you have a sec.

PR #264, opened 2018-01-05
 Updated the formatting script to have a default branch 

garethellis0:
@RobynCastro - I figured this is a good opportunity to learn some more about how the all the automated formatting stuff works. ~The only change here is the `fix_formatting.sh` file, the rest of the diff is because I branched off the changes from #259 to avoid future conflicts (that also means this PR is blocked on #259).~ Lemme know if you have questions about how the whole clang-format thing works.

### DIFF
--- a/clang-format/fix_formatting.sh
+++ b/clang-format/fix_formatting.sh
@@ -1,17 +1,34 @@
 #!/bin/bash
 
+# This script can be called with no arguments (`./fix_formatting`), in which 
+# case DEFAULT_BRANCH will be used as the branch to compare against to find 
+# changes to format, or it can be called with a specific branch to compare 
+# against (`./fix_formatting master`)
+
 # The version of the clang executable to use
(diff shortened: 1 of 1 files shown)


garethellis0:
Ah, so `OUTPUT` here is being set to the output of the command to the right. To verify this for yourself, you can run the following in a bash shell:
```
➜  ~ echo "Hello World"
Hello World
➜  ~ MY_VARIABLE="$(echo "Hello World")"
➜  ~ echo $MY_VARIABLE
Hello World
```
`CURR_DIR` is just a variable that holds the path to the folder that this script is in (not where it was run from). Beyond that, yeah, I'm not sure how it works either :grimacing:.

garethellis0:
Looking at the script, a better comment for the `CURR_DIR` variable would be helpful. I've updated that accordingly. Can you re-approve?

Issue #231, opened 2017-10-06
Create an Obstacle Manager


garethellis0:
This will act like a global map for us, general structure would be:

- take in `igvc_obstacle_list` message (note that this message type does not yet exist, see #228 )
- compare all the obstacles in the message to all already known obstacles, merge/update any that are similar enough
- purge all obstacles outside of a given radius of the robot (reducing issues with drift due to inaccurate sensor data)
- broadcast a `nav_msgs/OccupancyGrid` message with all the obstacles on it inflated by some given radius

**Note** 
There is a lot of complexity in the above structure that likely needs to be hashed out some more before we implement this. For example:
- how do we know when two obstacles are "similar"?
- how do we merge/update obstacles?

garethellis0:
After some thought (pictured below):

## Determining If Two Obstacles are Close together
- what does it mean to say that two obstacles are "close together"?
### Cones
- fairly easy, just need to check if `d = sqrt(d_x^2 + d_y^2)` is below some `d_max`
### Lines
- much harder
- initial plan (bottom right of first image) is to decompose both lines into a series of points and doing some sort of line of best through them, then checking if the sum error from the regression is below some `error_max`

## Merging Two Obstacles
- once we know that two obstacles are the "same", how do we merge them?
### Cones
- pretty easy, we probably just want to take the newer cone, since they are discrete and Lidar is pretty accurate
### Lines
- much harder
- initial plan (bottom of second image) is to decompose the lines into points and do a regression line through them. Priority can be given to the new or old lines (or equally), depending on what error the regression prioritizes minimizing.

![img_20171104_134631](https://user-images.githubusercontent.com/9075711/32409332-f3fedd64-c166-11e7-95ec-088e47faf023.jpg)
![img_20171104_134623](https://user-images.githubusercontent.com/9075711/32409335-f99160c6-c166-11e7-85ec-4eca118db1c9.jpg)

raadkhan:
After doing some searching/thinking about lines, (assuming the line segments are a vector of points) we could determine if two line segments are the "same" by seeing if two conditions are both true:
Condition 1: take dot product of the lines, that gives cosine of angle between them, if the value is above some threshold (close to 1) then we can declare them as "same" in terms of slope. 
Condition 2: compare how close is new line's start point to the old line's end point or if the line segments intersect, then we can determine if the lines are close enough to be merged.

garethellis0:
[...]

Issue #229, opened 2017-10-06
Create a Node to Extract Cones from a LaserScan


garethellis0:
Much of the work here was done in the `LidarObstacleManager` under the `iarrc` branch. The algorithms may stay the same, but it should be rewritten from scratch. Genercal Structure would be:

- take in a pointcloud
- transform the pointcloud from the lidar's frame of reference to the frame of reference of the robot
- group scan hits (2D points) within a given tolerance of each other
- create a "Cone" obstacle for each hit
- broadcast an `igvc_obstacle_list` message (note that this message type does not yet exist, see #228 )

garethellis0:
Note that when #273 is finished, we could use this instead of the Lidar, or use it to further validate whether or not a cone is actually present.

Issue #227, opened 2017-10-06
Create a Node to Extract Lines From PointClouds


garethellis0:
We'd like a node that can extract mathematical lines from pointclouds. The general structure here would be:
- take in the pointcloud
- group points that are within a given range of each other
- do a polynomial line of best fit through the points (we should be able to easily change the degree of the polynomial)
- broadcast an `igvc_obstacle_list` message (note that this message type does not yet exist, see #228 )

![image](https://user-images.githubusercontent.com/9075711/31260479-129ff922-aa03-11e7-880e-3a2c85d92ace.png)

garethellis0:
An interesting paper on extracting objects from 3D point clouds can be found here: https://arxiv.org/abs/1711.06396

mingyokim:
# My approach to this problem so far:
## Assumptions:

* The data points [(x,y,z),(h,s,v)] that do not pass the HSV filter are simply removed from PCL
* Bounding box is described in 2D

## Implementation:

* PCL is in 3D, but what we really want is a 2D projection from above. This can be achieved by ignoring the z component.  
* Remove points from PCL that are outside the bounding box.
* Perform density-based clustering on the resulting PCL. 
  * Rough idea for algorithm: [link](https://www.cs.ubc.ca/~schmidtm/Courses/340-F17/L9.pdf) (taken from my ML class)
  * This would theoretically work well if HSV filtering produces distinct groups of points for each line, but it would be quite sensitive when the boundaries between the groups are ambiguous.
  * We would have to fine-tune the hyper parameters for minimum neighbors and radius by observing its performance on sample data. 
  * We may also have an expected number of clusters (most likely 1 or 2 lines?); in the case where we have more clusters than we expected, we can just keep the clusters that contain the highest number of points and discard the rest. In the case where we have less clusters than we expected, we could re-run density-based clustering  with stricter parameters (for minimum neighbors and radius), but this is probably computationally inefficient. 
* Perform non-linear regression by means of change of basis (this can take in p for degree of polynomial so we can easily adjust this) for each cluster. However, keep in mind that the number of basis grows exponentially with the degree of polynomial (O(2^p)). If we want to use a higher degree of polynomial, we would need to use kernels but I'm assuming we would probably stay around p=3.

garethellis0:
[...]

PR #268, opened 2018-01-20
Created custom messages for representing IGVC obstacles

### LINKED ISSUE #228: Create an "igvc_obstacle_list" message type
Should contain a list of `obstacles`, where an `obstacle` is either (currently) a polynomial line of arbitrary degree, or a cone (a point). However, we should also be able to easily add obstacle types.

garethellis0:
I'm opening a pull request for these messages before the obstacle manager (#231) is done so that we establish what sorts of messages we'll be passing out of the line from pointcloud extractor (#227) and the cone extractor (#229). It also resolves #228 (ended up deciding to go with specific message types instead of generic single one just because of how ROS message types are created)

FYI @mingyokim - this contains the `LineObstacle` message type that you'll be passing out of the line extractor. Let me know if you need any help adding this message type to your package

@ValRat - I'd like @raadkhan to review this one before you do to give him a bit more experience with doing reviews from scratch


src/mapping_igvc/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+##     catkin_package(CATKIN_DEPENDS ...)
+##   * uncomment the add_*_files sections below as needed
+##     and list every .msg/.srv/.action file to be processed
+##   * uncomment the generate_messages entry below
+##   * add every package in MSG_DEP_SET to generate_messages(DEPENDENCIES ...)
+
+## Generate messages in the 'msg' folder
+add_message_files(
+    DIRECTORY msg
+    FILES ConeObstacle.msg LineObstacle.msg Point2D.msg
```
raadkhan: It looks like `LineObstacle.msg` is abstracting a piece-wise polynomial, implying that line obstacles are represented by single polynomial segments, and merging two line obstacles gives us a spline. So do we want a spline `msg` here as well?

src/mapping_igvc/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
+cmake_minimum_required(VERSION 2.8.3)
```
raadkhan: `CMakeLists` looks good as far as I can see

src/mapping_igvc/CMakeLists.txt
garethellis0: I considered it, but right now I don't see any need to publish the spines as messages. If all goes to plan, by the time they leave the obstacle manager, it's in the form of an occupancy grid.

raadkhan: This makes sense 👍

src/mapping_igvc/CMakeLists.txt
ValRat: Personally, I'm more of the form of nuking most if not all commented template entries.
Cleaner and easier to parse on first glance - but this is personal preference and I won't go on a murder spree for it :P

src/mapping_igvc/package.xml
```diff
@@ -0,0 +1,69 @@
+<?xml version="1.0"?>
```
ValRat: Same comment as the CMakelists.txt for removing templated entries

src/mapping_igvc/msg/Point2D.msg
```diff
@@ -0,0 +1,3 @@
+# A 2d point
```
ValRat: Would [`geometry_msgs/Pose2D`](http://docs.ros.org/api/geometry_msgs/html/msg/Pose2D.html) be appropriate as a substitute for this message?

src/mapping_igvc/package.xml
```diff
@@ -0,0 +1,69 @@
...
+<package format="2">
+  <name>mapping_igvc</name>
```
ValRat: Would it be worth the time to make the messages their own package? 
I'm assuming this current one will contain everything regarding mapping (obstacle manager, occ grid publisher, etc.)

src/mapping_igvc/CMakeLists.txt
garethellis0: Hmm, I think it's important that we leave at least some commented entries in for the sections that we are using. While I agree that it would be at least a bit cleaner and easier for us to work with, we should keep in mind that we want to keep this as accessible as possible for new members. For now, I think I'm inclined to keep the comment density as is, but I've asked for some feedback in the #software-leads channel on this, so we can at least have a discussion about it.

src/mapping_igvc/msg/Point2D.msg
garethellis0: I considered it, but the meaning of a [`geometry_msgs/Pose2D`](http://docs.ros.org/api/geometry_msgs/html/msg/Pose2D.html) is different from just a 2D point, and since this ends up inside another custom message anyways (`ConeObstacle`), this doesn't really add any complexity to either usage or the build process.

src/mapping_igvc/package.xml
garethellis0: Ok, roughly the same sort of reply here as with the `CMakeLists.txt`, I've made another note in the #software-leads channel. Once we decide on things there we can always come back and nuke some (or all) comments.

src/mapping_igvc/package.xml
garethellis0: I definitely considered this, and my initial setup for these message types was to put them in a separate package, just because the build stuff was a little easier. The balance here is that it logically makes sense for them to be in this package (I think), and reduces the number of packages in `src` (which hopefully keeps things a bit more accessible for new members), but does add a little bit of complexity to the build process (we need to make sure messages are built before executables) and breaks a ROS pseudo-standard (that messages have their own package). It won't take long to put these in a separate package, so if you feel at all strongly about this, reply as such and I can.

src/mapping_igvc/package.xml
garethellis0: After some feedback, looks like we should keep things as they are, leaving comments in for the portions of the `CMakeLists.txt` that we're using.

garethellis0:
@ValRat - comments addressed, can you re-review?

PR #269, opened 2018-01-22
Moving green light and circle detection from iarrc to snowflake

garethellis0:
So this was originally submitted as #224, but I've had to do some mildly terrifying things with git to convince that we're **not** trying to nuke the repo..... so I'm opening a new pull request.


garethellis0:
Alrighty, after a *little* more work then we were initially expecting, we've fixed all the major issues up, so this should be good for a review - @RobynCastro? (@ValRat - can you hold off on reviewing until Robyn does? Just trying to give some practice reviewing things from scratch)

src/sb_vision/src/CircleDetection.cpp
```diff
@@ -0,0 +1,141 @@
...
+    vector<vector<Point>> contours;
+    vector<Vec4i> hierarchy;
+    vector<cv::Point2i> center;
+    vector<float> radii;
+
+    // Convert to a grayscale image, if channels do not match.
+    cv::Mat bwImage;
+    switch (filtered_image.channels()) {
+        // 4 channel and 3 channel conversion should cover all our use cases.
+        case 4: cvtColor(filtered_image, bwImage, CV_BGRA2GRAY); break;
```
RobynCastro: Case statements looks so much better when it's just on one line.

garethellis0:
Alright, @ValRat - want to take a run at this one? (I had to update the branch to be in sync with master, so I need another review anyhow).

PR #274, opened 2018-02-05
Setup .gitignore .rosinstall .get_started.sh to install dependent packages

### LINKED ISSUE #272: New Phidget IMU Integration
Phidget IMU sponsored has its own USB interface. Motivation for switching so we can remove arduino. 
---
Product Page for IMU: https://www.phidgets.com/?tier=3&catid=4&pcid=2&prodid=1010
Product Page for Interface: https://www.phidgets.com/?tier=3&catid=4&pcid=2&prodid=1040
--- 
The interface has software documentation on the website.

Dulluhan:
This is to resolve #272 and the dependency of submodules
This should allow us to remove the git submodules.  
Files will install into ` /usr/share/ros/ `
I've tested things myself running the get_started.sh script myself, everything should work.  

---  
To manually do this: 
1. `sudo mkdir /usr/share/ros` can change if you have preferred location
2. `sudo chmod a+rwx /usr/share/ros` 
3. `rosinstall .` In the top level of our repo where the .rosinstall file is
4. source setup.bash 
--- 
Notes: don't `sudo rosinstall` it changes permissions of your .rosinstall and not sure if that will also affect ability to run installed packages.


.rosinstall
```diff
@@ -3,3 +3,12 @@
...
 # IF YOU CHANGE IT, USE rosinstall FOR THE CHANGES TO TAKE EFFECT
+- setup-file: {local-name: /opt/ros/kinetic/setup.sh}
+- git: {local-name: /usr/share/ros/phidgets_api, uri: 'https://github.com/ros-drivers-gbp/phidgets_drivers-release.git',
+    version: release/kinetic/phidgets_api/0.7.5-0}
+- git: {local-name: /usr/share/ros/sicktoolbox, uri: 'https://github.com/ros-drivers/sicktoolbox.git'}
+- git: {local-name: /usr/share/ros/sicktoolbox_wrapper, uri: 'https://github.com/ros-drivers/sicktoolbox_wrapper.git'}
+- git: {local-name: /usr/share/ros/zed-ros-wrapper, uri: 'https://github.com/stereolabs/zed-ros-wrapper.git'}
+- git: {local-name: /usr/share/ros/ros_arduino_bridge, uri: 'https://github.com/hbrobotics/ros_arduino_bridge.git'}
+- git: {local-name: /usr/share/ros/razor_imu_9dof, uri: 'https://github.com/KristofRobot/razor_imu_9dof.git'}
+- git: {local-name: /usr/share/ros/nmea_navsat_driver, uri: 'https://github.com/ros-drivers/nmea_navsat_driver.git'}
```
garethellis0: For several of these repositories, we have our own fork with a few changes. You should be able to the repo's we're using if you look in the `.gitmodules` file in the top level of the repo.

get_started.sh
```diff
@@ -75,6 +75,12 @@ sudo apt-get install ros-kinetic-desktop-full
 sudo rosdep init
 rosdep update
 
+# Run rosinstall for other packages
+sudo mkdir /usr/share/ros
```
garethellis0: I think this will fail (and kill this script) if the directory already exists. We should be using `sudo mkdir -p` instead.

garethellis0: What was the reasoning behind this path? I'd think we'd want to closely emulate the normal ros install path structure (so something more like `/usr/share/ros/kinetic/` (though obviously this is a bit different, as normally ros would install stuff to `/opt/ros/kinetic/share`, and we're already in a "share" folder).

get_started.sh
```diff
@@ -75,6 +75,12 @@ sudo apt-get install ros-kinetic-desktop-full
...
+sudo chmod a+rwx /usr/share/ros
+rosinstall .
+echo $DIR/setup.bash >> ~/.bashrc
```
garethellis0: If there are new lines we need to add to the `.bashrc` file, they should be added to the `new_shell_config_lines` array on line 26. This prevents duplicate lines from being added to users config files (checks if they're already present), and lets us support multiple shells (bash, zsh, fish, etc.)

garethellis0: I think we should probably be using the `setup.sh` script instead of `setup.bash`. In my experience this is the most universal option (ie. will usually work on bash and non-bash shells).

garethellis0: Also, I would expect that we would want to run `source $DIR/setup.sh` instead of just `$DIR/setup.bash`? (The difference being that `source` will cause the script to be run in the *current* shell, rather then spawning a new shell to run it)

.gitignore
```diff
@@ -19,3 +19,7 @@ cmake-build-debug
 *_REMOTE_*.txt
 *.orig
 
+setup.bash 
+setup.sh 
+setup.zsh
```
garethellis0: What do these setup files normally contain? Might it be more appropriate to put them in the same location that we're installing the packages to (similar in structure to how `/opt/ros/kinetic` is structured)?

get_started.sh
garethellis0: One additional comment, since this has to do with installing dependencies for the repo, I think that everything but adding the `$DIR/setup.bash` file to the `.bashrc` file should probably be in the `install_depdencies` script.

Dulluhan:
Looks like CI failed because of an apt problem? Seems that it couldn't connect to the repos

get_started.sh
Dulluhan: The first reason I chose this was to differentiate between what was installed via rosinstall and apt. The second reason being that I was worried the setup scripts would conflict with each other since they are setup the same way. If I could figure out a way to have rosinstall be smart enough to merge the setup scripts, then we could change the path to match. But we can add a share folder there, no big difference, just didnt see us needing other folders under `opt/ros/kinetic/` to have to go one dir deeper.

get_started.sh
Dulluhan: Comment 1: Will do
Comment 2: Sure
Comment 3: Forgot about that, will do
Comment 4: Ok

.rosinstall
Dulluhan: I'll take a look at our own repo

.gitignore
Dulluhan: Like mentioned, I haven't figured out how to merge the setup scripts cleanly yet.

Dulluhan: I will need to spend some time to actually look at the contents of the setup scripts or read some docs.

Dulluhan:
The CI integration passed because setup.sh was allowed_failure. Refer to yaml config file. We will have to decide in the long term if we would rather have rosinstall fail or have CI fail, because it currently looks like this is a conflict between wstool and rosinstall where rosinstall is looking for a setup.sh file location to know where the ros dist is installed, but wstool does not like anything in the `.rosinstall` file that is not a package link. This also affects subbots repo https://github.com/ubc-subbots/Jormungandr/pull/30

Dulluhan:
I will take a look to see if `wstool merge` has other configuration updates.

garethellis0:
Sorry, I'm still a bit confused what's going on here?
- CI is failing for this pull request. As far as I can can tell from the CI config, we're not letting `setup.sh` be an "allowed failure". Just to double check, which line in which file are you looking at?
- We definitely can't have CI or rosinstall fail, so we'll have to modify at least one of them 
- How did we sneak this past CI on `Jormungandr`, but CI is failing here?

Dulluhan:
So the persisting problem with travis will be addressed in #275. This PR was originally closed because of the branch is was pulling from. So I am going to re close this branch. All future references to Travis failing with rosinstall should be discussed and fixed in #275 or alternatively open a new thread to figure out how to fix this problem

PR #276, opened 2018-02-10
Quick README Cleanup

garethellis0:
Just a few small touch-ups to the README
- removed a dead link
- added a section for new members that links to `README_NEWMEMBERS.md`
- removed some old descriptions from when we had separate `iarrc` and `igvc` branches


garethellis0:
@raadkhan - merged those changes, should be good now

PR #277, opened 2018-02-11
Path to Twist (Issue 233)

WilliamGu98:
I completed a first implementation of an algorithm to convert a path message to a twist (also given the robot's current position in the global frame). I wrote a couple ROS tests and the algorithm is working as I expected, but I am unsure how to further test it to confirm that the it would work in practice.


src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,157 @@
...
+    float desired_angle = atan(y_sum/x_sum);
+    float current_angle = robotOrientation;
+
+    float turn_rate = desired_angle - current_angle;
+    float speed = 1.0 - fabs(fmod(turn_rate,2*M_PI)/(2*M_PI));
+
+    twist_msg.linear.x = speed;
+    twist_msg.angular.z = turn_rate;
+
+    return twist_msg;
```
garethellis0: In order to make testing easier, ideally we would have a single function `static` function that you can pass the current position/rotation of the robot and the path into, and which produces a  `Twist` message.

src/pathfinding_igvc/test/path-finding-test.cpp
```diff
@@ -0,0 +1,32 @@
...
+    //PathFinding test_node(NULL,NULL,NULL);
+
+    std::vector<float> testVec;
+    testVec.push_back(1);
+    testVec.push_back(2);
+    testVec.push_back(3);
+
+    EXPECT_EQ(1.0, PathFinding::weightedYSum(testVec, 1));
+}
+
```
garethellis0: Once we have the function I mentioned above, we can try to test:
- different starting orientation (what if we're perpendicular to the first vector in the path? what if we're facing the opposite direction?)
- different starting position (what if we're a meter away from the start of the path? we should make sure to have a few tests located far away from (0,0) to make sure that doesn't break anything)
- paths with immediate sharp left/right turns
- paths with a few consecutive vectors that point in radically different directions (ie. directly right, directly left, 45 degrees right, etc.)

garethellis0:
Just a final note that I'm really happy to see you asking questions like this! Testing like this is a super important way to catch bugs before we integrate the whole system, and it really helps the team when people make it a priority!

WilliamGu98:
Thanks! I'll be sure to make the changes you highlighted when I get the chance.

WilliamGu98:
I've made the changes that you requested and also added a few G-tests covering the different scenarios you outlined. The results of the tests seem correct but some numbers might still need scaling adjustments.

src/pathfinding_igvc/test/path-finding-test.cpp
```diff
@@ -0,0 +1,219 @@
...
+        poses.push_back(pose_stamped);
+    }
+    path_msg.poses = poses;
+
+    geometry_msgs::Twist twist_msg =
+    PathFinding::pathToTwist(path_msg, xPos, yPos, orientation);
+    std::cout << "Perpen Path" << std::endl;
+    std::cout << twist_msg.linear.x << std::endl;
+    std::cout << twist_msg.angular.z << std::endl;
+    std::cout << std::endl;
```
garethellis0: In general, we shouldn't really be printing anything in tests unless it directly relates to the test passing/failing (most of which gtest already tells us). This is to ensure that the output from full test runs (when we run all the tests), remains as compact and readable as possible.

src/pathfinding_igvc/test/path-finding-test.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+TEST(PathFinding, testWeightedSum2) {
+    std::vector<float> testVec;
+    testVec.push_back(1);
+    testVec.push_back(2);
+    testVec.push_back(3);
+
+    EXPECT_EQ(1.0, PathFinding::weightedSum(testVec, 1));
+}
+
```
garethellis0: A test for vectors with negative values might be another good case to consider here.

src/pathfinding_igvc/test/path-finding-test.cpp
```diff
@@ -0,0 +1,219 @@
...
+    PathFinding::pathToTwist(path_msg, xPos, yPos, orientation);
+    std::cout << "Erratic Path" << std::endl;
+    std::cout << twist_msg.linear.x << std::endl;
+    std::cout << twist_msg.angular.z << std::endl;
+    std::cout << std::endl;
+    EXPECT_NEAR(0.96, twist_msg.linear.x, 0.01);
+    EXPECT_NEAR(0.25, twist_msg.angular.z, 0.01);
+}
+
+// Case where position/orientation of robot is ahead of the path
```
garethellis0: I really like the descriptive comments in front of each test, makes it much easier to quickly read through them all and see what sorts of things we're testing.

src/pathfinding_igvc/test/path_finding_rostest.cpp
```diff
@@ -0,0 +1,133 @@
...
+    loop_rate.sleep();
+
+    // spinOnce allows ros to actually process your callbacks
+    // for the curious:
+http://answers.ros.org/question/11887/significance-of-rosspinonce/
+    ros::spinOnce();
+
+    EXPECT_FLOAT_EQ(0.875, speed);
+    EXPECT_FLOAT_EQ(M_PI/4, turn_rate);
+}*/
```
garethellis0: We should uncomment these ROSTests and make sure they're passing.

src/pathfinding_igvc/include/PathFinding.h
```diff
@@ -0,0 +1,54 @@
+//
+// Created by william on 20/01/18.
+//
+
+#ifndef PATHFINDING_IGVC_PATHFINDING_H
+#define PATHFINDING_IGVC_PATHFINDING_H
+
+#include "../../sb_utils/include/sb_utils.h"
```
ValRat: This can be cleaned up by referencing `sb_utils` in `CMakelists.txt` and `package.xml` and then instead:
```
#include "sb_utils.h"
```

WilliamGu98:
I've made changes according to all your comments.

src/pathfinding_igvc/CMakeLists.txt
```diff
@@ -0,0 +1,210 @@
...
+
+    # Adding rostest to the package
+    find_package(rostest REQUIRED)
+    # name the test and link it to the .test file and the .cpp file itself, this will allow
+    # "catkin_make run_tests" to be able to find and run this rostest
+    add_rostest_gtest(path_finding_rostest test/pathfinding_igvc_test.test test/path_finding_rostest.cpp)
+    target_link_libraries(path_finding_rostest ${catkin_LIBRARIES})
+endif()
+
+## Add folders to be run by python nosetests
```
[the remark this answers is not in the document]
WilliamGu98: Done

src/pathfinding_igvc/include/PathFinding.h
```diff
@@ -0,0 +1,54 @@
+//
+// Created by william on 20/01/18.
+//
```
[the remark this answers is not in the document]
WilliamGu98: Done

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+    robot_x_pos      = 0;
+    robot_y_pos      = 0;
+    robotOrientation = 0;
+
+    // Setup Subscriber to path
+    std::string path_subscribe_topic = "/path";
+    path_subscriber                  = nh.subscribe(
+    path_subscribe_topic, queue_size, &PathFinding::pathCallBack, this);
+
+    std::string tf_subscribe_topic = "/tf";
```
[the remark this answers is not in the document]
WilliamGu98: Done

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Produces twist message from a path message and current robot
+ * coordinates/orientation
+ * @param path_msg
+ * @param x_pos
+ * @param y_pos
+ * @param orientation
+ * @return a path message
+ */
```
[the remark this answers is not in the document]
WilliamGu98: Done

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+geometry_msgs::Twist PathFinding::pathToTwist(nav_msgs::Path path_msg,
+                                              double x_pos,
+                                              double y_pos,
+                                              double orientation) {
+    geometry_msgs::Twist twist_msg; // Initialize velocity message
+    const int NUM_POSES =
+    10; // Number of pose points to process including initial robot position
```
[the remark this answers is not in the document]
WilliamGu98: Done

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    float desired_angle = atan(y_sum / x_sum);
+    float current_angle = orientation;
+
+    float turn_rate = fmod(desired_angle - current_angle,
+                           2 * M_PI); // Keep turn_rate between -2pi and 2pi
+    if (turn_rate > M_PI)
+        turn_rate -= 2 * M_PI;
+    else if (turn_rate < -M_PI)
+        turn_rate += 2 * M_PI;
```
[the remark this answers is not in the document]
WilliamGu98: Didn't change because in a previous line I already fmodded the turn rate so that it should always be between -2pi and 2pi

src/pathfinding_igvc/test/path-finding-test.cpp
WilliamGu98: Done

src/pathfinding_igvc/test/path_finding_rostest.cpp
WilliamGu98: Done

src/pathfinding_igvc/include/PathFinding.h
WilliamGu98: Done

src/pathfinding_igvc/include/PathFinding.h
```diff
@@ -0,0 +1,108 @@
...
+ * Created By: William Gu
+ * Created On: Jan 20, 2018
+ * Description: A path finding node which converts a given path into a Twist
+ * message to send to the robot
+ */
+
+#ifndef PATHFINDING_IGVC_PATHFINDING_H
+#define PATHFINDING_IGVC_PATHFINDING_H
+
+//#include "../../sb_utils/include/sb_utils.h"
```
garethellis0: We should remove commented out code.

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,144 @@
...
+    SB_getParam(
+    private_nh, "global_frame", global_frame, (std::string) "odom_combined");
+    SB_getParam(private_nh, "num_poses", num_poses, 10);
+}
+
+void PathFinding::pathCallBack(const nav_msgs::Path::ConstPtr& path_ptr) {
+    nav_msgs::Path path_msg =
+    *path_ptr; // Take required information from received message
+    geometry_msgs::Twist twist_msg = pathToTwist(
+    path_msg, robot_x_pos, robot_y_pos, robotOrientation, num_poses);
```
garethellis0: What about the case where we receive a path before a tfcallback? Won't `robot_x_pos` and `robot_y_pos` still be 0?

src/pathfinding_igvc/test/path_finding_rostest.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    // Wait for the message to get passed around
+    ros::Rate loop_rate(1);
+    loop_rate.sleep();
+
+    ros::spinOnce();
+
+    EXPECT_NEAR(0.80, speed, 0.1);
+    EXPECT_NEAR(1.00, turn_rate, 0.1);
+}
```
garethellis0: A RosTest testing out publishing a few tf's would be good to have here.

src/pathfinding_igvc/test/path-finding-test.cpp
WilliamGu98: Fixed

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+ * in member variables
+ */
+void PathFinding::tfCallBack(const tf2_msgs::TFMessageConstPtr tf_message) {
+    // Check if the tf_message contains the transform we're looking for
+    for (geometry_msgs::TransformStamped tf_stamped : tf_message->transforms) {
+        // header frame might be base
+        if (tf_stamped.header.frame_id == global_frame &&
+            tf_stamped.child_frame_id == base_frame) {
+            // We've found the transform we're looking for, so see how close we
+            // are to the waypoint
```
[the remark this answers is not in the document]
WilliamGu98: Removed

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // Setup NodeHandles
+    ros::init(argc, argv, node_name);
+    ros::NodeHandle nh;
+    ros::NodeHandle private_nh("~");
+    uint32_t queue_size = 1;
+
+    /*Assume initial coordinates: may need to change later*/
+    robot_x_pos      = 0;
+    robot_y_pos      = 0;
+    robotOrientation = 0;
```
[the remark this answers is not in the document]
WilliamGu98: Yeah, these values are just placeholders for now

src/pathfinding_igvc/src/PathFinding.cpp
```diff
@@ -0,0 +1,188 @@
...
+            double y_pos = tf_stamped.transform.translation.y;
+
+            double quatx = tf_stamped.transform.rotation.x;
+            double quaty = tf_stamped.transform.rotation.y;
+            double quatz = tf_stamped.transform.rotation.z;
+            double quatw = tf_stamped.transform.rotation.w;
+            tf::Quaternion q(quatx, quaty, quatz, quatw);
+            tf::Matrix3x3 m(q);
+            double roll, pitch, yaw;
+            m.getRPY(roll, pitch, yaw);
```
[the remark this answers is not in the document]
WilliamGu98: Fixed

src/pathfinding_igvc/src/PathFinding.cpp
WilliamGu98: I was thinking of adding some type of flag variable that's set to true once we receive the first tfcallback and then checking the flag for each pathcallback so that a twist message won't be published if the flag is false. I don't know if there's a better way to do this though, since we would need to check the flag for every pathcallback.

src/pathfinding_igvc/test/path_finding_rostest.cpp
WilliamGu98: Added tf publishers to ROS tests with some additional cases for testing them specifically

src/pathfinding_igvc/src/PathFinding.cpp
garethellis0: That seems like a pretty reasonable approach to me!

garethellis0:
@WilliamGu98 - is this good to go for a round of review?

WilliamGu98:
Yes please

Issue #271, opened 2018-01-27
Enable Olaf to brake


garethellis0:
It would be nice to have the ability to brake by reversing the current through the motor to induce resistance. This capability should be added in firmware

**TODO**
- Measure stopping distance currently
- Implement Braking
- Measure new stopping distance and report findings here

raadkhan:
Can reopen if we decide to ressurect Olaf

Issue #250, opened 2017-11-04
Improve the Twist Message -> Actual Movement Mapping in Olaf Firmware


garethellis0:
Currently, we just have an arbitrary relationship between the `angular_z` and `linear_x` and the actual turn rate and motor speed. `angular_z` is supposed to be in rad/s and `linear_x` is in m/s. 

There are two main points to be done here:
- implement some sort of relation between `angular_z` and `angle` under the `drive` function, assuming `angular_z` is in rad/s, such that we turn at the expected rate (up to maximum the servo turning the wheels can go)
- implement some sort of relation between `linear_x` and `velocity` under the `drive` function, assuming `linear_x` is in m/s, such that we move at the expected rate

marinah1:
@garethellis0 Pushed firmware onto https://github.com/marinah1/Snowflake

raadkhan:
Can reopen if we decide to ressurect Olaf

PR #279, opened 2018-02-17
Removed a old version of the control firmware for Olaf

garethellis0:
This was the version of the firmware that `olaf_control` is based off of. It's now defunct (replaced by `olaf_control`).


garethellis0:
Marinah will be working on this, see #250 and #271

PR #283, opened 2018-03-01
Simplifying Travis CI

garethellis0:
After banging on CI for a *considerable* amount of time (see commit history) below, I've come up with a (hopefully) simpler and easier to understand CI system. Migrating to this will allow us to remove our dependecy on `moveit_ci` and delete the [UBC-Snowbots/moveit_ci](https://github.com/UBC-Snowbots/moveit_ci/) repo.


Dulluhan:
Can you merge master into your branch first

setup_scripts/install_arduino.sh
```diff
@@ -0,0 +1,49 @@
...
+echo "Linking Arduino"
+sudo rm /usr/local/bin/arduino
+sudo ln -s -f $ARDUINO_INSTALL_DIR/$ARDUINO_DIR_NAME/arduino /usr/local/bin/arduino
+
+echo "Applying Snowbots modifications to core Arduino files"
+
+## Add the `flushRX` function. For context, see issue #176
+# Add the implementation to the `.cpp` file
+# Remove the last line (the final `#endif`) so we're inside the guards
+sed -i '$ d' $ARDUINO_INSTALL_DIR/$ARDUINO_DIR_NAME/hardware/arduino/avr/cores/arduino/HardwareSerial.cpp
```
Dulluhan: Function declaration in the header file?

setup_scripts/install_tools.sh
```diff
@@ -59,22 +59,6 @@ echo "================================================================"
 # Add the user to the dialout so they can do arduino things
 sudo adduser $USER dialout
```
Dulluhan: Probably want to move this line into the arduino setup file

setup_scripts/install_tools.sh
```diff
@@ -110,7 +94,7 @@ cd $DIR
 # Setup Snowbots Udev Rules  #
 ##############################
 cd $DIR
-./src/firmware/setup_udev_rules.sh
+./../src/firmware/setup_udev_rules.sh
```
Dulluhan: I think we should move all the script files out of src, maybe place setup_udev_rules.sh into this $DIR. Another thought is the udev setup for phidget there are two ways to look at it. 1. Move all the udev rules into one file. 2. Move all the udev rules to their respective scripts. (i.e If udev setup_udev_rules.sh is for arduino, then move udev setup into the arduino install script)

travis_ci.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+
+
+# Change to the directory this script is in
+cd $CURR_DIR
+
+# Note that we must build the codebase in order to run tests
+if [ "$RUN_BUILD" == "true" ] || [ "$RUN_TESTS" == "true" ]; then
+    # Install all required dependecies
+    travis_run ./setup_scripts/install_dependencies.sh
```
Dulluhan: ~~I just checked the install_dependencies.sh, rosinstall command is missing, iirc rosinstall should also be run in this script~~

Dulluhan: NVM, we can do this later, I can add it in through the other PR once we merge this one. Makes more sense that way. However it makes my previous comment about udev ruels awkward because where do we put phidget udev mods...

setup_scripts/install_arduino.sh
Dulluhan: Could we actually have someone look at if that is necessary, maybe it was just me being religious about it (flushRX). If the we don't actually drop messages, then we should remove this probably, unless someone has also noticed this fixing the problem. Who worked on this last year?

Dulluhan:
I assume you've tried running Travis on a docker image already?

setup_scripts/install_arduino.sh
garethellis0: Good catch, I just went through all the firmware code and it looks like we're no longer using the `flushRX` function, so I've removed this section of the setup script.

setup_scripts/install_tools.sh
garethellis0: I like the idea of having a unified udev rules file and having that file and the script to install it under `setup_scripts`. I've moved the `setup_udev_rules.sh` and `10-snowbots.rules` into `setup_scripts`.

travis_ci.sh
garethellis0: This should still work I think. `setup_udev_rules.sh` is called from `install_tools.sh`, *not* `install_dependencies.sh`, so CI never calls it. I'd advocate for having the phidget rules go in `setup_scripts/10-snowbots.rules`

garethellis0:
I've addressed/replied to all the comments. Thanks for the feedback!

What do you mean by "tried running Travis on a docker image"? Currently, our CI system already runs in docker, and this pull request is already running the updated CI version, feel free to take a look at the CI logs to see if anything looks off.

garethellis0:
I haven't tried a totally clean run on a standalone docker image (not in CI), but CI runs everything but the tools setup, and I just re-ran that on my computer and everything seems good to go!

PR #295, opened 2018-03-27
Line extractor rviz enhancement

mingyokim:
A couple of fixes and improvements regarding LineExtractorNode and rviz:
- Make a point cloud generating node that generates a point cloud that changes randomly over time
- Fix the bug where the clusters don't change when the point cloud changes over time
- Make `frame_id` configurable from launch file
- Display lines as series of points instead of line strips (using line strips requires using `MarkerArray`, but I've been encountering issues where some line strips in `MarkerArray` are not deleted)


src/sb_pointcloud_processing/test/TestUtils.h
```diff
@@ -47,8 +47,14 @@ class TestUtils {
     static void addLineToPointCloud(LineArgs args,
```
garethellis0: We should probably add a Javadoc style function comment here.

src/sb_pointcloud_processing/src/test_pcl_generator_node.cpp
```diff
@@ -100,20 +105,20 @@ sensor_msgs::PointCloud2 generatePclMessage(bool include_outlier) {
     // Generate a single PointCloud with noise
     pcl::PointCloud<pcl::PointXYZ> pcl;
     LineExtractor::TestUtils::addLineToPointCloud(
-    args, pcl, max_noise_x, max_noise_y);
+    args, pcl, max_noise_x, max_noise_y, false);
```
garethellis0: In general, when writing tests we want the results to be "deterministic" (that is, that same every time). if I understand the `same_seed` argument correctly, setting it false means that the line generated will be generated differently each time. One option to make this deterministic might be to add another argument to `addLineToPointCloud` that lets you set the seed used to generate the line, then manually choose a seed and set `same_seed` to `false`.

garethellis0:
@mingyokim - is this ready for re-review? If so, could you quickly reply to the two review comments?

src/sb_pointcloud_processing/src/test_pcl_generator_node.cpp
mingyokim: done

src/sb_pointcloud_processing/test/TestUtils.h
mingyokim: done

mingyokim:
@garethellis0 it's ready!

src/sb_pointcloud_processing/src/test_pcl_generator_node.cpp
```diff
@@ -100,25 +108,29 @@ sensor_msgs::PointCloud2 generatePclMessage(bool include_outlier) {
...
     sensor_msgs::PointCloud2 msg;
     pcl::toROSMsg(pcl, msg);
 
     return msg;
-}
\ No newline at end of file
+}
+
+unsigned int getSeed() {
+    return counter++ % 10;
```
garethellis0: We should add a javadoc comment for this function. Also, why are we `% 10`'ing this? won't this mean we will only ever generate 10 seeds?

mingyokim: Yes, the intention was to visually (because it's rviz) ensure that the test works for 10 seeds. If we have `MAX_INT` number of seeds, we can't guarantee that the test works for all of them unless we wait/watch long enough

PR #296, opened 2018-03-28
Line extractor performance enhancement

mingyokim:
This is an early PR and not finalized. The goal is to get early feedback/ideas on how to enhance performance. Also, this branch is based off of #295 which is not merged yet, so only look at commits starting from "initial implementation of multithreading".


mingyokim:
This PR is now ready for review (starting from the commit "initial implementation of multithreading").

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -7,6 +7,8 @@
 #ifndef LINE_EXTRACTOR_IGVC_DBSCAN_H
 #define LINE_EXTRACTOR_IGVC_DBSCAN_H
 
+#define SEQUENTIAL_CUT_OFF 5000
```
garethellis0: What was the reasoning behind making this a pre-processor definition? If possible we should make this a settable class member

garethellis0: A comment explaining what this is used for would also be really helpful.

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -78,6 +100,36 @@ class DBSCAN {
...
+    /**
+     * Finds the neighbours in parallel
+     * Number of threads is defined by this->_num_threads
+     */
+    void findNeighborsParallel();
+
+    /**
+     * Finds the neighbours sequentially
+     */
+    void findNeighborsSequential();
```
garethellis0: What was the reasoning behind keeping `findNeighboursParallel` and `findNeighboursSequential`? I would think they would be equivalent if we just set `num_threads` to 1?

src/sb_pointcloud_processing/src/DBSCAN.cpp
```diff
@@ -98,10 +109,92 @@ void DBSCAN::findNeighbors() {
...
+            std::cerr << "failed to start worker" << std::endl;
+        }
     }
+
+    // join threads
+    for (unsigned int i = 0; i < this->_num_threads; i++) {
+        if (pthread_join(threads[i], NULL)) {
+            std::cerr << "failed to join worker" << std::endl;
+        }
+    }
```
garethellis0: It's not quite clear how we don't just exit here? Maybe make a note about what the threads are blocking on?

src/sb_pointcloud_processing/src/DBSCAN.cpp
```diff
@@ -98,10 +109,92 @@ void DBSCAN::findNeighbors() {
         }
 
         // store the list of neighbours
-        this->_neighbors.insert({i, neighbors});
+        this->_neighbors[i] = neighbors;
+    }
+}
```
garethellis0: Just out of curiosity, how does doing this with pthreads compare to using omp? Something like:
```
void DBSCAN::findNeighborsSequential() {
    #pragma omp parallel for 
    for (unsigned int i = 0; i < this->_pcl.size(); i++) {
        vector<unsigned int> neighbors;
        pcl::PointXYZ current_point = this->_pcl[i];

        // for current_point, determine all neighbour points that are within
        // predetermined radius
        for (unsigned int j = 0; j < this->_pcl.size(); j++) {
            if (i == j) continue;
            pcl::PointXYZ neighbor_point = this->_pcl[j];
            if (dist(current_point, neighbor_point) <= this->_radius) {
                neighbors.push_back(j);
            }
        }

        // store the list of neighbours
        this->_neighbors[i] = neighbors;
    }
}
```
If this works, I feel like it might be a bit cleaner
(for reference: https://bisqwit.iki.fi/story/howto/openmp/#ExampleInitializingATableInParallelMultipleThreads)

mingyokim: confirmed that this works just fine

src/sb_pointcloud_processing/include/DBSCAN.h
mingyokim: I made the sequential cut off a pre-processor definition because it's far more unlikely that we would be constantly changing this value unlike some of the other sensitive hyperparameters we have for DBSCAN like `radius` and `min_neighbours`. Also "standard" values for sequential cut off ranges from 1000 to 5000. That was my reasoning, let me know if you still think we should make this a configurable value

src/sb_pointcloud_processing/include/DBSCAN.h
mingyokim: using openmp instead, not relevant anymore

src/sb_pointcloud_processing/src/DBSCAN.cpp
mingyokim: using openmp instead, not relevant anymore

mingyokim:
@garethellis0 it's ready for re-review!

src/sb_pointcloud_processing/include/DBSCAN.h
garethellis0: Hmm, I see what you mean, but I don't think it'd hurt to make it a class member, at the very least it might be a bit more clear that this is part of the class itself.

src/sb_pointcloud_processing/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY VALUE "Release")
 endif()
 
-add_definitions(-std=c++14)
+add_definitions(-std=c++14 -fopenmp)
```
garethellis0: I'm not sure what happens if we have a compiler flag for a feature that isn't present on the computer. A better way to add OpenMP might be the following (which would go below with the rest of the `find_package` lines):
```
find_package(OpenMP REQUIRED)
if(OPENMP_FOUND)
message("OPENMP FOUND")
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}")
endif()
```
cmake doc reference: https://cmake.org/cmake/help/v3.0/module/FindOpenMP.html

I think this would also mean you wouldn't have to link against `gomp` for all the executables (the compiler would just have universal openmp support). Normally linking something to every executable is bad practice, but in this case the compiler is literally just replacing your code with a multithreaded version before it compiles it (the actual multithreading is done with `pthread`, so you *might* have to link against that), so there's no performance loss.

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -7,6 +7,9 @@
 #ifndef LINE_EXTRACTOR_IGVC_DBSCAN_H
 #define LINE_EXTRACTOR_IGVC_DBSCAN_H
 
+// run sequentially when point cloud size is less than SEQUENTIAL_CUT_OFF
+#define SEQUENTIAL_CUT_OFF 1000
```
garethellis0: I replied to your comment on the previous review, but including it in this review for ease of reading:

"Hmm, I see what you mean, but I don't think it'd hurt to make it a class member, at the very least it might be a bit more clear that this is part of the class itself."

src/sb_pointcloud_processing/src/DBSCAN.cpp
```diff
@@ -83,6 +80,9 @@ void DBSCAN::expand(unsigned int center_index,
 }
 
 void DBSCAN::findNeighbors() {
+    this->_neighbors = new vector<unsigned int>[this->_pcl.size()];
+
+#pragma omp parallel for if (this->_pcl.size() > SEQUENTIAL_CUT_OFF)
```
garethellis0: Glad to see this worked! This is really easy to read now. A quick comment explaining what this line is doing would be good for people who've never worked with pre-compiler directives / openmp before (I'm also a tad confused on how this `if` option works, I've never used it before).

garethellis0: Just another note, out of curiosity, did we see any performance change after the switch to openmp?

mingyokim: They were very negligible, if any.

src/sb_pointcloud_processing/CMakeLists.txt
mingyokim: done

src/sb_pointcloud_processing/include/DBSCAN.h
mingyokim: done. just to clarify, do we want to make this a variable that we can configure from the launch file?

mingyokim:
@garethellis0 ready for review!

src/sb_pointcloud_processing/include/DBSCAN.h
garethellis0: From your previous comments, it doesn't seem like this is something that we should need to change (if I understand correctly)? In this case it's more about making it clear that it's part of the class and can be changed if need be, so I don't see a need to make it a rosparam as well at this point.

src/sb_pointcloud_processing/CMakeLists.txt
```diff
@@ -81,6 +89,7 @@ target_link_libraries(line_extractor_node
     ${catkin_LIBRARIES}
     ${PCL_COMMON_LIBRARIES}
     ${PCL_IO_LIBRARIES}
+    gomp
```
garethellis0: Just noticed this, do we still need to link against `gomp`?

mingyokim: nope. removing it

mingyokim:
@garethellis0 could you review once again? thanks!

PR #298, opened 2018-03-29
Fixed a few small things in the sample_package

garethellis0:


src/sample_package/src/MyNode.cpp
```diff
@@ -15,18 +15,19 @@ MyClass::MyClass(int argc, char **argv, std::string node_name) {
...
     std::string topic_to_subscribe_to = "subscribe_topic";
-    int refresh_rate = 10;
-    my_subscriber = nh.subscribe(topic_to_subscribe_to, refresh_rate, &MyClass::subscriberCallBack, this);
+    int queue_size = 10;
+    my_subscriber = nh.subscribe(topic_to_subscribe_to, queue_size, &MyClass::subscriberCallBack, this);
 
     // Setup Publisher(s)
     std::string topic = private_nh.resolveName("publish_topic");
-    uint32_t queue_size = 1;
+    queue_size = 10;
```
RobynCastro: isn't this a redundant line? but otherwise looks good!

garethellis0: It is redundant, but I thought it best to set it again here so people don't get the idea that the queue sizes need to be the same. I'll change the value so it's a bit more clear

garethellis0:
In this case the queue size really doesn't matter. But in general: https://answers.ros.org/question/190357/rospypublisher-whats-a-good-choice-for-queue_size/

garethellis0:
@RobynCastro - ready for re-review

PR #301, opened 2018-04-08
WIP: Created a node to translate the output from the Phidget Encoders to Odometry Readings

garethellis0:
**Summary of changes**
- cleaned up the `drivers` package a bit, especially the `CMakeLists.txt` and `package.xml` files
- added a launch file for the phidget encoders
- created/tested a node to translate phidget encoder output into an [Odometry msg](http://docs.ros.org/api/nav_msgs/html/msg/Odometry.html)

**Note:** Still a work in progress, a few TODO's left to address, but it's close enough that I'd like to request a review now so we can get this merged in ASAP

@Dulluhan for review on covariance generation, specifically: `calculatePoseCovariance`
@ValRat for general review
@mingyokim for general review

### DIFF
--- a/src/drivers/launch/config/test_cart.yaml
+++ b/src/drivers/launch/config/test_cart.yaml
@@ -1,0 +1,22 @@
+# The radius of the wheels
+wheel_radius: 0.13
+
+# The distance between the wheels
+wheelbase_length: 0.66
+
+# The number of ticks for a single rotation of a wheel
+# This will depend on both the number of ticks the encoder reads 
+# and the gear ratio between the encoders and the wheels
+ticks_per_rotation: 3600
+
+# If set to `true` this node assumes that *positive* encoder
+# ticks mean the wheel is moving *forward*. If the encoders are 
+# mirrored, it is likely one of them will need to be reversed
+reverse_left_encoder: false
+reverse_right_encoder: true
+
+# The variance of the left and right encoders.
+# TODO: This could use a more comprehensive comment about what variance is and how to determine it 
+left_encoder_variance: 0.1
+right_encoder_variance: 0.1
+
--- a/src/drivers/launch/config/test_cart.yaml
+++ b/src/drivers/launch/config/test_cart.yaml
@@ -1,3 +1,5 @@
+# This file defines physical parameters for the test cart (built for 2017-2018)
+
 # The radius of the wheels
 wheel_radius: 0.13
 
--- a/src/drivers/launch/encoder_odometry_node.launch
+++ b/src/drivers/launch/encoder_odometry_node.launch
@@ -3,12 +3,14 @@
 
     <!-- Start the Encoder to Odometry Node -->
     <node name="encoder_odometry_node" pkg="drivers" type="encoder_odometry_node" output="screen" >
-        <param name="left_encoder_joint_name" value="left_encoder" />
-        <param name="right_encoder_joint_name" value="right_encoder" />
-        <param name="wheel_radius" value="0.1" />
-        <param name="wheelbase_length" value="0.5" />
-        <param name="ticks_per_rotation" value="1000" />
-        <param name="odom_msg_refresh_rate" value="5" />
+        <rosparam param="left_encoder_joint_name"> "left_encoder" </rosparam>
+        <rosparam param="right_encoder_joint_name"> "right_encoder" </rosparam>
+        <rosparam param="reverse_left_encoder"> false </rosparam>
+        <rosparam param="reverse_right_encoder"> true </rosparam>
+        <rosparam param="wheel_radius"> 0.13 </rosparam>
+        <rosparam param="wheelbase_length"> 0.66 </rosparam>
+        <rosparam param="ticks_per_rotation"> 3600 </rosparam>
+        <rosparam param="odom_msg_refresh_rate"> 10 </rosparam>
     </node>
 
 </launch>
--- a/src/drivers/launch/encoder_odometry_node.launch
+++ b/src/drivers/launch/encoder_odometry_node.launch
@@ -1,0 +1,14 @@
+
+<launch>
+
+    <!-- Start the Encoder to Odometry Node -->
+    <node name="encoder_odometry_node" pkg="drivers" type="encoder_odometry_node" output="screen" >
+        <param name="left_encoder_joint_name" value="left_encoder" />
+        <param name="right_encoder_joint_name" value="right_encoder" />
+        <param name="wheel_radius" value="10" />
+        <param name="wheelbase_length" value="0.5" />
+        <param name="ticks_per_rotation" value="14000" />
+        <param name="odom_msg_refresh_rate" value="5" />
+    </node>
+
+</launch>
--- a/src/drivers/launch/encoder_odometry_node.launch
+++ b/src/drivers/launch/encoder_odometry_node.launch
@@ -5,9 +5,9 @@
     <node name="encoder_odometry_node" pkg="drivers" type="encoder_odometry_node" output="screen" >
         <param name="left_encoder_joint_name" value="left_encoder" />
         <param name="right_encoder_joint_name" value="right_encoder" />
-        <param name="wheel_radius" value="10" />
+        <param name="wheel_radius" value="0.1" />
         <param name="wheelbase_length" value="0.5" />
-        <param name="ticks_per_rotation" value="14000" />
+        <param name="ticks_per_rotation" value="1000" />
         <param name="odom_msg_refresh_rate" value="5" />
     </node>
 
--- a/src/drivers/launch/encoder_odometry_node.launch
+++ b/src/drivers/launch/encoder_odometry_node.launch
@@ -3,14 +3,31 @@
 
     <!-- Start the Encoder to Odometry Node -->
     <node name="encoder_odometry_node" pkg="drivers" type="encoder_odometry_node" output="screen" >
+        <!-- 
+            Encoder Ticks should be published as a JointState massage,
+            this defines what names in the JointState correspond to the left 
+            and right encoders
+        -->
         <rosparam param="left_encoder_joint_name"> "left_encoder" </rosparam>
         <rosparam param="right_encoder_joint_name"> "right_encoder" </rosparam>
-        <rosparam param="reverse_left_encoder"> false </rosparam>
-        <rosparam param="reverse_right_encoder"> true </rosparam>
-        <rosparam param="wheel_radius"> 0.13 </rosparam>
-        <rosparam param="wheelbase_length"> 0.66 </rosparam>
-        <rosparam param="ticks_per_rotation"> 3600 </rosparam>
+
+        <!-- 
+            Number of times per second to generate a new Odometry message.
+            Higher values are not neccesarily better, as enough time should 
+            be allowed between Odometry estimates so that we are estimating
+            a new state based on several encoder readings
+        -->
         <rosparam param="odom_msg_refresh_rate"> 10 </rosparam>
+
+        <!-- The frame the Odometry messages will be generated in -->
+        <rosparam param="odom_frame"> "encoder" </rosparam>
+
+        <!-- 
+            Load the physical parameters that define the physical robot
+            see examples in the `config` folder.
+        -->
+        <rosparam command="load" 
+            file="$(find drivers)/launch/config/test_cart.yaml" />
     </node>
 
 </launch>
--- a/src/drivers/launch/phidget_encoders.launch
+++ b/src/drivers/launch/phidget_encoders.launch
@@ -5,8 +5,6 @@
     <node name="encoders_driver" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen">
         <rosparam param="joint0_name">left_encoder</rosparam>
         <rosparam param="joint1_name">right_encoder</rosparam>
-        <rosparam param="joint0_tick2rad">1</rosparam>
-        <rosparam param="joint1_tick2rad">1</rosparam>
         <remap from="/joint_states" to="/encoders/joint_states" />
     </node>
 
--- a/src/drivers/launch/phidget_encoders.launch
+++ b/src/drivers/launch/phidget_encoders.launch
@@ -1,11 +1,13 @@
 
 <launch>
 
-    <!-- Load which Phidget's we're using (via their serial numbers) -->
-    <rosparam command="load" file="$(find drivers)/config/phidget_encoders_test_cart.yaml" />
-
-    <!-- Start the encoder nodes -->
-    <node name="left_encoder" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen" />
-    <node name="left_encoder" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen" />
+    <!-- Start the encoder node that gets encoder ticks over serial -->
+    <node name="encoders_driver" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen">
+        <rosparam param="joint0_name">left_encoder</rosparam>
+        <rosparam param="joint1_name">right_encoder</rosparam>
+        <rosparam param="joint0_tick2rad">1</rosparam>
+        <rosparam param="joint1_tick2rad">1</rosparam>
+        <remap from="/joint_states" to="/encoders/joint_states" />
+    </node>
 
 </launch>
--- a/src/drivers/launch/phidget_encoders.launch
+++ b/src/drivers/launch/phidget_encoders.launch
@@ -1,0 +1,11 @@
+
+<launch>
+
+    <!-- Load which Phidget's we're using (via their serial numbers) -->
+    <rosparam command="load" file="$(find drivers)/config/phidget_encoders_test_cart.yaml" />
+
+    <!-- Start the encoder nodes -->
+    <node name="left_encoder" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen" />
+    <node name="left_encoder" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen" />
+
+</launch>
--- a/src/drivers/src/EncoderOdometryNode.cpp
+++ b/src/drivers/src/EncoderOdometryNode.cpp
@@ -19,7 +19,8 @@ EncoderOdometryNode::EncoderOdometryNode(int argc, char **argv, std::string node
     left_encoder_num_ticks_curr({}),
     right_encoder_num_ticks_curr({}),
     left_encoder_num_ticks_prev({}),
-    right_encoder_num_ticks_prev({})
+    right_encoder_num_ticks_prev({}),
+    last_estimate(generateAllZeroOdometryMessage())
 {
     // Setup NodeHandles
     ros::init(argc, argv, node_name);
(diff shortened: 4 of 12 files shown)


src/drivers/src/encoder_odometry_node.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+    // Set the refresh rate to the desired odom estimation rate
+    ros::Rate loop_rate(odometry_estimate_refresh_rate);
+
+    // Keep making estimates until we're told to stop
+    while (ros::ok()){
+        encoder_odometry_node.publishEstimatedOdomMsg();
+        ros::spinOnce();
+        loop_rate.sleep();
+    }
```
ValRat: Could consider using a [Timer](http://wiki.ros.org/roscpp/Overview/Timers) within the node itself perhaps?

src/drivers/test/encoder_odometry_node_rostest.cpp
```diff
@@ -0,0 +1,227 @@
...
+ */
+class EncoderOdometryNodeTest : public testing::Test{
+protected:
+
+    EncoderOdometryNodeTest():
+            wheel_radius(0.1),
+            ticks_per_rotation(1000),
+            wheelbase(0.5)
+            {
+        // TODO: Delete me
```
ValRat: Delete me :D

src/drivers/test/encoder_odometry_node.test
```diff
@@ -0,0 +1,19 @@
+<!-- ROSTests for the encoder_odometry_node -->
+
+<launch>
+
+    <!-- Start the Encoder to Odometry Node -->
+    <!-- TODO: THis should *not* output to "screen" -->
```
ValRat: Is this `TODO` still relevant?
Also noticed that `output="screen"` is also present in `encoder_odometry_node.launch` as well.

src/drivers/src/EncoderOdometryNode.cpp
```diff
@@ -0,0 +1,277 @@
...
+    // TODO: Figure out actually reasonable variances
+    SB_getParam(private_nh, "left_encoder_variance", left_encoder_variance, 0.1);
+    SB_getParam(private_nh, "right_encoder_variance", right_encoder_variance, 0.1);
+
+    // Setup Subscriber(s)
+    joint_state_subscriber = nh.subscribe(std::string("/encoders/joint_states"), 10, &EncoderOdometryNode::encoderJointStateCallback, this);
+    reset_subscriber = private_nh.subscribe(std::string("reset"), 1, &EncoderOdometryNode::resetCallback, this);
+
+    // Setup Publisher(s)
+    std::string odom_estimate_topic_name = nh.resolveName("/encoders/odom");
```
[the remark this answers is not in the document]
garethellis0: That was the intent here. I thought it might be a bit cleaner if all the encoder "stuff" (joint_states, Odometry), was under the same namespace. I've added an extra note to the code for now, but let me know if you think we ought to put it under a local namespace.

src/drivers/test/encoder_odometry_node_rostest.cpp
garethellis0: :laughing: - deleted

src/drivers/src/encoder_odometry_node.cpp
garethellis0: Huh, didn't know this was a thing, TIL! Anyhow, my initial thought was that it might be a bit harder to follow then the current structure (using `while(ros::ok())`) but after some more consideration this seems like a much better solution. People on the team are already used to having basically all the functionality in the class itself, and are already used to the concept of callback functions, so this isn't much of a leap in understanding. Changed!

src/drivers/test/encoder_odometry_node.test
garethellis0: Welp, glad you're catching all the TODO's I missed lol. The reason I didn't want this one to output to screen is that it's a test, and so we want to keep the test logs clean so that it's easy to see/find failures when running lots of tests (for example in CI). I don't think there's too much issue with having normal launch files outputting to `screen`. Changed: removed from `encoder_odometry_node.test` but **not** from `encoder_odometry_node.launch`

garethellis0:
@ValRat - changes made, ready for re-review!

src/drivers/launch/encoder_odometry_node.launch
```diff
@@ -0,0 +1,33 @@
...
+
+        <!-- The frame the Odometry messages will be generated in -->
+        <rosparam param="odom_frame"> "encoder" </rosparam>
+
+        <!-- 
+            Load the physical parameters that define the physical robot
+            see examples in the `config` folder.
+        -->
+        <rosparam command="load" 
+            file="$(find drivers)/launch/config/test_cart.yaml" />
```
Dulluhan: Is there a way for the config files to be loaded dynamically instead of having to change the launch file? If not, could we change the launch file to match the config file?

src/drivers/launch/phidget_encoders.launch
```diff
@@ -0,0 +1,11 @@
+
+<launch>
+
+    <!-- Start the encoder node that gets encoder ticks over serial -->
+    <node name="encoders_driver" pkg="phidgets_high_speed_encoder" type="phidgets_high_speed_encoder" output="screen">
+        <rosparam param="joint0_name">left_encoder</rosparam>
```
Dulluhan: Document the numbering somewhere so everyone knows? Minor, but at some point we should have a way of referring to all sensors esp duplicated ones to match the indexing number we give them. Eg. From the back left to right side of robot is always 0 index start incrementing to the right. Minor, just thinking for confused future people.

src/drivers/src/EncoderOdometryNode.cpp
```diff
@@ -0,0 +1,290 @@
...
+    covariances[0*6 + 1] = cov_x_y;
+    covariances[0*6 + 5] = cov_x_theta;
+    covariances[1*6 + 0] = cov_y_x;
+    covariances[1*6 + 1] = cov_y_y;
+    covariances[1*6 + 2] = cov_y_theta;
+    covariances[5*6 + 0] = cov_theta_x;
+    covariances[5*6 + 1] = cov_theta_y;
+    covariances[5*6 + 5] = cov_theta_theta;
+
+    return covariances;
```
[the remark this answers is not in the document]
Dulluhan: Further inspection, I think that this node is overlapping with the functionality of the EKF node. Awks. The paper you found was used to compare against EKF in performance. Though in comparison, it seems that due to the linearization in the EKF, it is not good at adapting angle changes that are large. Imagine inverse tangential plot. So in theory this filter should be better. But it does not take in to account other sensors. Perhaps the best way to approach this problem is to merge the covariance calculations. Off the top of my head, our covariances for this calculation is convoluted while the covariances with EKF will be linear, because assumed independent events. 

TL;DR - This is overlapped functionality with EKF. I don't think we should integrate the covariance calculations here into the EKF, but figure out how to use it on its own. I will need to think some more about this. Sorry.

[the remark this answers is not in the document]
Dulluhan: Clarifications: 

- The paper you use also assume independence between sensors.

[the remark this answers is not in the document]
Dulluhan: We can probably try to be cheeky and incorporate their pose estimation into our EKF and modify the covariance estimations to be nonlinear as well.

src/drivers/src/EncoderOdometryNode.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // TODO: Detailed doc comments here
+    // let p_t be the change from the previous state (x,y,theta): p_t = [x_t; y_t; theta_t]
+
+    // Cov[p_t] = E[p_t * transpose(p_t)] - E[p_t] * E[transpose(p_t)]
+
+    // Compute E[p_t * transpose(p_t)] - E[p_t]
+    double k_1 = std::pow(d_r, 2) + std::pow(d_l, 2);
+    double k_2 = std::pow(d_r, 2) - std::pow(d_l, 2);
+    double k_3 = d_r * d_l;
+    double k_4 = left_encoder_variance + right_encoder_variance;
```
Dulluhan: Reality is that all we needed to do was to provide `k_4` and `k_5`

src/drivers/launch/config/test_cart.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+# If set to `true` this node assumes that *positive* encoder
+# ticks mean the wheel is moving *forward*. If the encoders are 
+# mirrored, it is likely one of them will need to be reversed
+reverse_left_encoder: false
+reverse_right_encoder: true
+
+# The variance of the left and right encoders.
+# TODO: This could use a more comprehensive comment about what variance is and how to determine it 
+left_encoder_variance: 0.1
```
Dulluhan: I personally believe that unless we are going past the max read speed, or approaching it, there should not be much variance coming from the sensor itself. If there was any error it would be from rounding and calculations done by the driver and the usb interface, since calculating quadrature should not be tooo bad?

Dulluhan: the only other case is slipping or mechanical failure, which at that point was supposed to be handled by the EKF anyways.

src/drivers/launch/encoder_odometry_node.launch
garethellis0: Hmm, I don't think there is a way to load them dynamically. I definitely agree that this is a less then ideal solution though. What do you mean by "change the launch file"? Change the name to match? I'm a bit against changing the name, if only because it means we'll need to duplicate this launch file for every robot. We could add some higher level launch files that set the config file and then run most of the stuff in this launch file (and then have one of _those_ for each robot), but that might increase launch file structure complexity as well. @ValRat - could I get your thoughts on this as well?

src/drivers/launch/phidget_encoders.launch
garethellis0: Good point. Added a quick comment to this launch file.

src/drivers/src/EncoderOdometryNode.cpp
```diff
@@ -0,0 +1,290 @@
...
+geometry_msgs::PoseWithCovariance::_covariance_type EncoderOdometryNode::calculatePoseCovariance(
+        int left_wheel_ticks,
+        int right_wheel_ticks,
+        double prev_yaw) {
+
+    // All math taken from: http://digital.csic.es/bitstream/10261/30337/1/Onto%20computing.pdf
+    // namely pages 1342 and 1343
+
+    // Compute displacement of the left and right wheels (relative to last estimate)
+    double wheel_circumference = 2 * M_PI * wheel_radius;
```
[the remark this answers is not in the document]
garethellis0: I definitely considered it, but seeing as `wheel_radius` is already a class member, I didn't want to duplicate information. ie, if `wheel_circumference` is a class member and we update `wheel_radius` somewhere, we now also have to make sure that we're updating `wheel_circumference` as well.

src/drivers/src/EncoderOdometryNode.cpp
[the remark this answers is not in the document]
garethellis0: Darn. Ok, we may have to nuke much of this anyways. Val and I did some testing, and it looks like `robot_pose_ekf` (which it looks like we might be using unless the custom EKF gets done soon) doesn't work if there are any numbers close to 0 in the covariance :cry:

src/drivers/src/EncoderOdometryNode.cpp
garethellis0: Sorry, I'm not sure I totally understand? How would that propagate through the rest of the calculations?

src/drivers/launch/config/test_cart.yaml
garethellis0: So does this mean the variance on the encoders should be 0? Sorry, bit confused.

garethellis0:
@ValRat - comments addressed, ready for re-review whenever you have some "free" time
@Dulluhan - comments all replied too, bit confused on a few things (probably mostly me just not understanding fully how the sensor error model should work)

src/drivers/launch/config/test_cart.yaml
Dulluhan: Basically, I don't think there should should be much variance, unless proven otherwise empirically

src/drivers/src/EncoderOdometryNode.cpp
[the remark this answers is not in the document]
Dulluhan: Basically what happens when you don't have covariance is that they would be considered independent events. Therefore mathematically speaking, when we perform the update step in our EKF our Kalman Gain will be Minimized and the updated data will be ignored.

The other way to think about it is that, since they're independent events, Kalman Filter will basically scrap the data that round and not use the data to make predictions that round.

src/drivers/src/EncoderOdometryNode.cpp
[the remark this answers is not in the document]
Dulluhan: Wheel radius in the config file?

src/drivers/launch/encoder_odometry_node.launch
Dulluhan: I mean, we would have to edit this launch file in any case where we want to use a different config yaml file. Is there maybe someway of doing it using the roslaunch command by feeding it a second paramater?

Dulluhan: So quick test shows that rosparam loaded variables are static until roscore dies. Maybe we can have people load the config files themselves? I know its one extra step, but imo it might be easier. Also loading other config files will just overwrite current rosparams anyways.

src/drivers/src/EncoderOdometryNode.cpp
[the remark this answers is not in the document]
garethellis0: Hmm, ok, but it doesn't publish an update to the state at all (I suppose because it doesn't have any reliable odometry data?)

src/drivers/src/EncoderOdometryNode.cpp
garethellis0: Do you mean we would give them to the kalman filter directly?

src/drivers/launch/encoder_odometry_node.launch
garethellis0: Hmm, I'd prefer to automate as much as possible (the end goal being a single command to run as much of the system as possible), as I think it will really reduce simple avoidable errors, but I'll let @ValRat take the final call on this one.

src/drivers/src/EncoderOdometryNode.cpp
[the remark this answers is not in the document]
garethellis0: Sorry, I meant that if we ever go to make some changes to this class that (for some reason) modify the `wheel_radius` (within the class) after it (and presumably `wheel_circumference`) are set, we would have to make sure that we're updating `wheel_circumference` as well. Basically I think it's added complexity without any major readability/performance benefits (though I definitely understand where you're coming from on this one).

Dulluhan:
I'll draft up a write up for explaining everything. So slight hold on comments from me.

garethellis0:
Ready for re-review, specifically:

@Dulluhan - wheel encoder odometry variance, both how we're setting it in the node, and the variance calculation procedure in `src/drivers/README.md`
@ValRat - general review

src/drivers/README.md
```diff
@@ -1,4 +1,10 @@
...
+
+1. Start the `encoder_to_odometry` node, and push the robot in a curve with a radius of at least 1 meters (the larger the better, just make sure you are pushing it in the exact same curve every time). Repeat this at least 10 times, re-starting the node each time, and recording the end `linear.x`, `linear.y`, `angular.z`, and `angular.w` values. (**Note:** The orientation in the `Odometry` msg is a quaternion. To get `theta` in radians, use an [online quaternion-radian converter](http://quaternions.online/) or use the math [from wikipedia](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Source_Code_2)).
+TODO: Photo of path here
+2. Calculate the *standard deviation* for `linear.x`, `linear.y`, and `angular.z`. You can do this by putting them into a spreadsheet and calculating `STDEV` for a column (see photo below). Take the square root of each of these values to get the *variance*.
```
Dulluhan: variance is standard deviation squared, not the other way around.

src/drivers/README.md
```diff
@@ -1,4 +1,10 @@
...
+## Determining Variance for encoder_to_odometry node
+The `linear.x`, `linear.y`, and `angular.z` in the `Odometry` msg that we publish based on encoder data all have associated variances that are stored along the diagonal of the `pose.covariance` matrix (for full documentation, see [Odometry msg documentation](http://docs.ros.org/api/nav_msgs/html/msg/Odometry.html)). These are set as parameters on the `encoder_to_odometry` node. These values are determined experimentally, and are different for every robot. Here are the steps to determine them:
+
+1. Start the `encoder_to_odometry` node, and push the robot in a curve with a radius of at least 1 meters (the larger the better, just make sure you are pushing it in the exact same curve every time). Repeat this at least 10 times, re-starting the node each time, and recording the end `linear.x`, `linear.y`, `angular.z`, and `angular.w` values. (**Note:** The orientation in the `Odometry` msg is a quaternion. To get `theta` in radians, use an [online quaternion-radian converter](http://quaternions.online/) or use the math [from wikipedia](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Source_Code_2)).
```
Dulluhan: Euler angles are yaw, pitch, roll angles, its the fundamental set of angles to get into any orientation in space you want. In our case, is it correct to be looking at conversion from Euler Angles to Quaternions? Or do we want to be looking at something else?

Dulluhan: Basically there is a mixture of notation here. `linear.x,y` are used in quaternions (euclidean space), `angular.z` is a Euler angle (generally coming in the whole set of `angular.x,y,z`, or around an oriented vector), whilst `angular.w` is a vector angle used in the quaternion ROS data type to describe a rotation. 

What I'm saying is, are you sure you are recording `angular.z` AND `angular.w` at the same time? Because by Euler's rotation rule, `angular.w` will encompass `angular.z`.

src/drivers/README.md
garethellis0: Good catch, thanks! I think this was really throwing us off....

src/drivers/README.md
garethellis0: I'm not sure I totally understand what you're getting at here? I was just trying to match the instructions with what someone looking at the output of `rostopic echo` would see (ie. `linear.x`, `linear.y`, `angular.w`, `angular.z`)? Do you not need `angular.z` and `angular.w` to calculate the yaw of the robot?

Dulluhan: `angular.z` is yaw of the robot. `angular.w` is the combination of yaw, pitch and roll on a chosen axis. Which is generally your x,y,z vector in quaternions.

garethellis0: Sorry, I still don't understand. I thought that you needed both the `z` and `w` components of a quaternion (which is what angular is) to get the yaw?

garethellis0: Ok, that makes sense, but I'm still not exactly sure what you're getting at here? Just to make sure we're on the same page, `linear.x,y,z` describe where we are as a point in 3D space (and I don't believe is in any way related to our orientation), and `angular.x,y,z,w` describe our orientation in 3D space in the form of a quaternion?

Dulluhan: It is `angular x,y,z` or `angular.w` chose one. Not both. They represent the same thing. Standard quaternion should be `angular.w`. `angular x,y,z` is not quaternion notation.

garethellis0: Ah, I think I understand. Here I'm just trying to follow ROS notation as much as possible to make it easy to follow (which does use `x,y,z,w` for quaternion representation, see http://docs.ros.org/api/geometry_msgs/html/msg/Quaternion.html).

garethellis0: Ahh, ok, I think I understand now, changes made, please re-review.

PR #306, opened 2018-04-28
Sensor and Kalman Filter Launch Files

garethellis0:
**Stalled on #301 - should be a lot simpler once it's merged**

Setup the launch files for running all the sensors and the kalman filter.


.rosinstall
```diff
@@ -1,19 +1,35 @@
+# Drivers for interfacing with GPS's that give us nmea sentences (like the swift GPS)
```
ValRat: Thanks for the description, but these get nuked when using the CLI commands to set/update repositories.
i.e. running `wstool set src/external_pkgs/some_pkg --git https://github.com/abc/def.git` will basically reset the `.rosinstall` to only contain the entries themselves. Could make it a req to only update these by hand, but that's a little weirdly restrictive :/

ValRat:
Looks good, gotta test out on the physical robot tomorrow to double check that everything runs but I don't expect anything to go wrong.

garethellis0:
Rip, too bad. I've removed the comments to prevent possible confusion in the future.

src/localisation_igvc/launch/gps_to_odom.launch
```diff
@@ -6,10 +6,10 @@
     -->
     <node name="gps_to_odom" pkg="gps_common" type="utm_odometry_node"> 
         <param name="rot_covariance" value="99999" /> 
-        <param name="frame_id" value="base_link" /> 
+        <param name="frame_id" value="gps" /> 
  
-        <remap from="odom" to="/gps_driver/odom" /> 
-        <remap from="fix" to="/gps_driver/navsatfix" /> 
+        <remap from="odom" to="/gps/odom" /> 
+        <remap from="fix" to="/gps/navsatfix" />
```
ValRat: On second look, the `nmea_gps.launch` file publishes in `/gps_driver/navsatfix` no? Why are we changing this?

ValRat: Nvm, found out we changed it to this in our fork.

garethellis0:
@ValRat - this will need re-approval when CI passes again (had to merge with master)

PR #309, opened 2018-05-03
Cone manager issue229

WilliamGu98:
This is what I have so far for the cone extractor node. There's some comments that still need to be removed and maybe more documentation and minor stuff to be added, but the implementation works for the tests I've written so far.
I would have added more tests myself but I used my own helper class to generate laserscan messages so I was wondering if it'd be better to just move on to physical testing as it'd give a better idea of what works or doesn't in this implementation. (the laserscan messages I'm generating might be "cleaner" than what we would see with the actual laser scan)
Also -- I found a circle fitting algorithm online that I've put in the implementation, that's more accurate than the previous algorithm I was using by simply generating a circle based on 3 points at a time from the edge points of a cone. I was wondering if it'd be okay to use it though since I'm not completely sure how it works myself and essentially just copied the code.


src/sb_laserscan_processing/src/ConeIdentification.cpp
```diff
@@ -0,0 +1,293 @@
+/**
+ * Created by William Gu on 24/03/18
+ * Implementation of Cone Identification methods for identifying cones from a laser msg
+ */
+
+#include <ConeIdentification.h>
+
+
+std::vector<mapping_igvc::ConeObstacle> ConeIdentification::identifyCones(const sensor_msgs::LaserScan &laser_msg, double dist_tol, double radius_exp, double radius_tol, int line_point_dist, double ang_threshold){
```
garethellis0: This function is pretty long (>100), I think it would be much more readable if we could break it up into functions.

src/sb_laserscan_processing/src/ConeIdentification.cpp
```diff
@@ -0,0 +1,293 @@
...
+                std::vector<std::vector<mapping_igvc::Point2D>> split_edges = splitEdge(edge_points, line_point_dist, ang_threshold); //Split edge points if multiple cones detected (Temp values right now as params)
+                for (int i = 0; i < split_edges.size(); i++){
+                    
+                    mapping_igvc::ConeObstacle potential_cone = edgeToCone(split_edges[i]);
+
+                    //Testing
+                    /*
+                    for (int j=0 ;j < split_edges[i].size(); j++){
+                        std::cout<<"("<<split_edges[i][j].x<<","<<split_edges[i][j].y<<")"<<std::endl; //Print points in edge
+                    }*/
```
garethellis0: In general, we should try and remove commented out code and/or testing code (though I understand this is a WIP).

ValRat:
Just a warning, the `Point2D` message @WilliamGu98 used has been moved into `sb_geom_msgs` in @garethellis0 Obstacle Manager Part 1/2 PR. The dependency is a bit broken atm.

ValRat:
Fix would require replacing all `mapping_igvc::Point2D` with `sb_geom_msgs::Point2D`

garethellis0:
@WilliamGu98 - don't worry about the issue mentioned above for now, we'll see which PR gets merged first then deal with it

src/sb_laserscan_processing/test/cone-identification-test.cpp
```diff
@@ -0,0 +1,514 @@
...
+    edge_points.push_back(p1);
+    edge_points.push_back(p2);
+    edge_points.push_back(p3);
+    edge_points.push_back(p4);
+    edge_points.push_back(p5);
+    edge_points.push_back(p6);
+
+    double slope = ConeIdentification::getRegressionSlope(edge_points);
+    EXPECT_NEAR(1.217, slope, 0.02);
+}*/
```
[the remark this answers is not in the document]
WilliamGu98: Fixed

PR #312, opened 2018-05-08
Path Finding: Part 2

mingyokim:
basically just copy pasted https://www.geeksforgeeks.org/a-search-algorithm/ for the A* algorithm...!

### DIFF
--- a/src/pathfinding_igvc/include/AStar.h
+++ b/src/pathfinding_igvc/include/AStar.h
@@ -20,8 +20,31 @@ class AStar {
     struct GridPoint {
         int col;
         int row;
-        GridPoint(int c = 0, int r = 0) : col(c), row(r){};
+        GridPoint(int c = 0, int r = 0) : col(c), row(r) {};
     };
+
+    typedef std::pair<double, std::pair<int, int>> GridPointWithScore;
+
+    struct Cell
+    {
+        // parent of cell
+        GridPoint parent = GridPoint(-1,-1);
+        // f = g + h
+        double f = FLT_MAX, g = FLT_MAX, h = FLT_MAX;
+    };
+
+    int _num_rows;
+    int _num_cols;
+
+    GridPoint _start;
+    GridPoint _goal;
+
+    std::vector<int8_t> _grid;
+
+    Cell **_cell_details;
+    bool **_closed_list;
+
+    std::set<GridPointWithScore> _open_list;
 
     /**
      * Takes an occupancy grid as well as start and goal points, and calculates
@@ -33,9 +56,20 @@ class AStar {
      * @return points stacked in order, where the top contains the starting
      * GridPoint and the bottom contains the goal GridPoint
      */
-    std::stack<GridPoint> run(nav_msgs::OccupancyGrid occupancy_grid,
+    static std::stack<GridPoint> run(nav_msgs::OccupancyGrid occupancy_grid,
                               GridPoint start,
                               GridPoint goal);
+    std::stack<GridPoint> aStarSearch();
+    bool isValid(int row, int col);
+    bool isUnBlocked(int row, int col);
+    bool isDestination(int row, int col);
+    double calculateHValue(int row, int col);
+    std::stack<GridPoint> tracePath();
+    bool processSuccessor(GridPoint successor, GridPoint parent);
+
+private:
+    AStar(nav_msgs::OccupancyGrid occupancy_grid, GridPoint start,
+          GridPoint goal);
 };
 
 #endif // PATHFINDING_IGVC_ASTAR_H
--- a/src/pathfinding_igvc/include/AStar.h
+++ b/src/pathfinding_igvc/include/AStar.h
@@ -128,9 +128,31 @@ private:
      */
     std::stack<GridPoint> aStarSearch();
 
+    /**
+     * Returns whether or not a point is inside the occupancy grid
+     *
+     * @param point
+     * @return true if point is inside the grid, false otherwise
+     */
     bool isValid(AStar::GridPoint point);
+
+    /**
+     * Returns whether or not a point on grid is blocked
+     *
+     * @param point
+     * @return true if point on grid is GRID_FREE, false if GRID_OCCUPIED
+     */
     bool isUnBlocked(AStar::GridPoint point);
+
+    /**
+     * Returns whether or not a point on grid is the destination
+     * (i.e. this->_goal)
+     *
+     * @param point
+     * @return true if point == this->_goal, false otherwise
+     */
     bool isDestination(AStar::GridPoint point);
+    
     double calculateHValue(AStar::GridPoint point);
     std::stack<GridPoint> tracePath();
     bool processSuccessor(GridPoint successor, GridPoint parent);
--- a/src/pathfinding_igvc/include/AStar.h
+++ b/src/pathfinding_igvc/include/AStar.h
@@ -152,9 +152,38 @@ private:
      * @return true if point == this->_goal, false otherwise
      */
     bool isDestination(AStar::GridPoint point);
-    
+
+    /**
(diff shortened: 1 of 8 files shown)


src/pathfinding_igvc/include/AStar.h
```diff
@@ -14,15 +15,81 @@
...
+     */
+    int _num_rows;
+    int _num_cols;
+
+    /*
+     * parameters for storing the starting and goal
+     * points in the grid
+     */
+    GridPoint _start;
+    GridPoint _goal;
```
garethellis0: A lot of these class members seem like they should be private? Is there any reason to have them be public?

src/pathfinding_igvc/include/AStar.h
```diff
@@ -33,9 +100,91 @@ class AStar {
...
+     *
+     * The constructor is private because a path would only be found once
+     * for a given occupancy grid and start and goal points.
+     *
+     * @param occupancy_grid occupancy grid
+     * @param start
+     * @param goal
+     */
+    AStar(nav_msgs::OccupancyGrid occupancy_grid, GridPoint start,
+          GridPoint goal);
```
garethellis0: I'm a little confused why this constructor is private?

src/pathfinding_igvc/src/AStar.cpp
```diff
@@ -6,10 +6,224 @@
...
+        this->_closed_list[i] = new bool[this->_num_cols];
+    }
+}
 
 std::stack<AStar::GridPoint> AStar::run(nav_msgs::OccupancyGrid occupancy_grid,
                                         GridPoint start,
                                         GridPoint goal) {
-    // TODO Part 2
+    return AStar(occupancy_grid, start, goal).aStarSearch();
+}
```
garethellis0: This interface is a bit confusing for me. Why don't we just use the Constructor and then call `aStarSearch` on the object (basically what you're doing here, just not sure why we have a function for it and a private constructor)?

src/pathfinding_igvc/src/AStar.cpp
```diff
@@ -6,10 +6,224 @@
...
+
+std::stack<AStar::GridPoint> AStar::aStarSearch() {
```
garethellis0: `aStarSearch` is a bit redundant (since we're calling it on the `AStar` class). I'd suggest `search` or `findPath`

src/pathfinding_igvc/src/AStar.cpp
```diff
@@ -6,10 +6,224 @@
...
+        /*
+        Generating all the 8 successor of this cell
+
+            N.W   N   N.E
+              \   |   /
+               \  |  /
+            W----CellDetail----E
+                 / | \
+               /   |  \
+            S.W    S   S.E
```
garethellis0: Loving the ASCII art :+1: , but the center section looks like it's a few spaces to the right, and the bottom section looks like it's one space to the right.

src/pathfinding_igvc/test/astar-test.cpp
```diff
@@ -0,0 +1,65 @@
+//
+// Created by min on 20/05/18.
+//
```
garethellis0: All files should have standard start-of-file comment

src/pathfinding_igvc/test/astar-test.cpp
```diff
@@ -0,0 +1,65 @@
...
+    mapMetaData.origin = origin;
+
+    /* OccupancyGrid */
+    // initialize occupancy grid
+    nav_msgs::OccupancyGrid grid;
+
+    // set mapMetaData
+    grid.info = mapMetaData;
+    grid.data = {
+            AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_OCCUPIED, AStar::GRID_OCCUPIED, AStar::GRID_OCCUPIED, AStar::GRID_FREE, AStar::GRID_OCCUPIED, AStar::GRID_OCCUPIED, AStar::GRID_FREE,
```
garethellis0: I really like this test, I think adding the following to the start of the file might make this diagram easier to read

```
typedef AStart::GRID_FREE _;
typedef AStar::GRID_OCCUPIED X;
```
And then changing the diagram accordingly

src/pathfinding_igvc/test/path-finder-test.cpp
```diff
@@ -8,6 +8,148 @@
...
+
+    /* OccupancyGrid */
+    // initialize occupancy grid
+    nav_msgs::OccupancyGrid grid;
+
+    // set mapMetaData
+    grid.info = mapMetaData;
+    grid.data = {
+            AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_OCCUPIED, AStar::GRID_OCCUPIED, AStar::GRID_OCCUPIED, AStar::GRID_FREE, AStar::GRID_OCCUPIED, AStar::GRID_OCCUPIED, AStar::GRID_FREE,
+            AStar::GRID_FREE, AStar::GRID_OCCUPIED, AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_OCCUPIED, AStar::GRID_FREE, AStar::GRID_FREE, AStar::GRID_FREE,
```
garethellis0: Similar comment to above about a potential way to make these sorts of tests easier to read.

src/pathfinding_igvc/src/AStar.cpp
mingyokim: btw I didn't make this... LOOL it was taken from geeksforgeeks.com lmao

src/pathfinding_igvc/src/AStar.cpp
mingyokim: If we were to make the constructor public, that means that we can create an `AStar` object and reuse it for finding path multiple times. First of all, I don't see any use to create an `AStar` object and find the path more than once with the exact same occupancy grid and starting/goal point. Second of all, even if we were to reuse the `AStar` object, we would need to ensure that all the instance variables in this class are reinitialized. If we just make the constructor private, we don't need to worry about the `AStar` object being created and reused multiple times without all of its instance variables being reinitialized. 

I actually had this issue with DBSCAN and the bug was discovered pretty late since it only happened after the first callback. So making it private would avoid any issue of this sort

Let me know what you think/if you have any better suggestions!

src/pathfinding_igvc/include/AStar.h
mingyokim: already commented below, but copy pasting here:

If we were to make the constructor public, that means that we can create an AStar object and reuse it for finding path multiple times. First of all, I don't see any use to create an AStar object and find the path more than once with the exact same occupancy grid and starting/goal point. Second of all, even if we were to reuse the AStar object, we would need to ensure that all the instance variables in this class are reinitialized. If we just make the constructor private, we don't need to worry about the AStar object being created and reused multiple times without all of its instance variables being reinitialized.

I actually had this issue with DBSCAN and the bug was discovered pretty late since it only happened after the first callback. So making it private would avoid any issue of this sort

Let me know what you think/if you have any better suggestions!

src/pathfinding_igvc/test/astar-test.cpp
mingyokim: wow good idea

src/pathfinding_igvc/include/AStar.h
garethellis0: OK, I follow. Let's just make sure to put a (condensed) version of your explanation in a comment in the code itself so that people in the future understand why we're doing this.

garethellis0:
LGTM, :+1: :tada: - just need to fix formatting to make CI happy, ping me after and I can re-approve.

garethellis0:
Looks like we need to merge in master again

PR #318, opened 2018-05-12
Path Finding: Part 1

mingyokim:
First PR for Path Finding node. Probably only going to be two more PR's after this.

This PR does not include the implementation of AStar nor the Path Finding node. Its only concern is taking an occupancy grid (nav_msgs/OccupancyGrid) as well as the starting point and goal point (geometry_msgs/Point) in map frame, then producing the shortest path (nav_msgs/Path).

The main function that the to-be-built Path Finding Node will be interacting with is `PathFinder::perform`.

*NOTE: All files named PathFinding will be renamed to something like PathToTwist by @WilliamGu98*


src/pathfinding_igvc/include/OccupancyGridResizeService.h
```diff
@@ -0,0 +1,32 @@
...
+     * below
+     * the initial origin of the grid.
+     * If the cell location is already inside the grid, it doesn't change the
+     * grid
+     *
+     * @param grid the occupancy grid passed by reference
+     * @param goal the cell location that we want to fit inside the grid
+     */
+    void static resizeOccupancyGridToFitGoal(nav_msgs::OccupancyGrid& grid,
+                                             AStar::GridPoint goal);
```
garethellis0: This seems like it might be a very expensive operation. Keep in mind we might have a very high resolution graph (say 10cm per point) that's 5m by 5m (25m^2) and the goal might be 50m away, in which case an occupancy grid of that size would probably use GB's of memory and take a while to build. 

My suggestion for this would be to increase the size of the occupancy grid by one in every direction (with all added grid cells being unoccupied) then make the goal node the  closest node in these additional nodes the closest node in these newly added nodes to the goal node
screenshot from 2018-05-16 21-57-48


Let me know what you think, this is definitely a very tricky problem, but I _think_ this solution should be fairly simple to implement?

src/pathfinding_igvc/include/PathConstructionService.h
```diff
@@ -0,0 +1,41 @@
...
+ *              containing poses in map frame
+ */
+
+#ifndef PATHFINDING_IGVC_PATHCONSTRUCTIONSERVICE_H
+#define PATHFINDING_IGVC_PATHCONSTRUCTIONSERVICE_H
+
+#include <AStar.h>
+#include <OccupancyGridConversionService.h>
+
+class PathConstructionService {
```
garethellis0: Suggest `PathConstructionService` -> `PathConstructor`

src/pathfinding_igvc/include/PathFinderTestUtils.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * Created By: Min Gyo Kim
+ * Created On: May 12th 2018
+ * Description: Class that provides helper functions for path-finder-test
+ */
```
garethellis0: This file should probably be in `src/pathfinding_igvc/test`?

src/pathfinding_igvc/src/OccupancyGridResizeService.cpp
```diff
@@ -0,0 +1,52 @@
+/*
+ * Created By: Min Gyo Kim
+ * Created On: May 14th 2018
+ * Description: Implementation of occupancy grid resize service
+ */
+
+#include <OccupancyGridResizeService.h>
+
+void OccupancyGridResizeService::resizeOccupancyGridToFitGoal(
+nav_msgs::OccupancyGrid& grid, AStar::GridPoint goal) {
```
garethellis0: We should add some comments to this function

src/pathfinding_igvc/include/PathFinder.h
```diff
@@ -0,0 +1,41 @@
...
+     * path
+     * from start to goal based on information from occupancy grid.
+     * @param start starting point in world frame
+     * @param goal goal point in world frame
+     * @param grid occupancy grid
+     * @return shortest path from start to goal
+     */
+    static nav_msgs::Path perform(geometry_msgs::Point start,
+                                  geometry_msgs::Point goal,
+                                  nav_msgs::OccupancyGrid grid);
```
garethellis0: `perform` isn't super descriptive, perhaps something more like `calculatePath`?

src/pathfinding_igvc/test/path-finder-utils-test.cpp
```diff
@@ -0,0 +1,29 @@
+/*
+ * Created By: Min Gyo Kim
+ * Created On: May 14th 2018
+ * Description: Unit tests for path finder utils class
+ */
```
garethellis0: this file should probably be called `pathfinder-test-utils.cpp`, `path-finder-utils-test` makes it sound like we're testing the utils

src/pathfinding_igvc/include/OccupancyGridResizeService.h
mingyokim: Yeah that would be much more efficient. Sounds easy to implement as well

src/pathfinding_igvc/src/OccupancyGridResizeService.cpp
mingyokim: probably going to rewrite this implementation based on earlier comment

src/pathfinding_igvc/test/path-finder-utils-test.cpp
mingyokim: This file is actually testing `PathFinderUtils`

garethellis0: Ah, that makes sense, all good.

mingyokim:
@garethellis0 ready for re-review!

src/pathfinding_igvc/include/PathFinderUtils.h
```diff
@@ -0,0 +1,126 @@
...
+     */
+    static tf::Quaternion getQuaternionFromAngle(double angle) {
+        tf::Quaternion q;
+        tf::Matrix3x3 rotationMatrix = tf::Matrix3x3();
+        rotationMatrix.setEulerYPR(
+        angle, 0.0, 0.0); // only set Z rotation since it's 2D
+        rotationMatrix.getRotation(q);
+
+        return q;
+    }
```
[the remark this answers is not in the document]
mingyokim: hmm one of my unit tests fails if I replace it with that

```
[ RUN      ] PathFinderUtils.TestGetAngleBetweenPoints
/home/min/Snowbots/Snowflake/src/pathfinding_igvc/test/path-finder-utils-test.cpp:25: Failure
Value of: q.getAngle()
  Actual: 2.6179938
Expected: 3.14159265358979323846 + 3.14159265358979323846 / 6
Which is: 3.6651914
[  FAILED  ] PathFinderUtils.TestGetAngleBetweenPoints (0 ms)
```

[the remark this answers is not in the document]
mingyokim: should be either `-2.6179938` or `3.6651914` but it seems like it's having some issues with negative angles

mingyokim:
@garethellis0 ready for re-re-review!

src/pathfinding_igvc/include/PathFinderUtils.h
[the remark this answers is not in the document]
garethellis0: Hmm, odd, but if your function works we can leave it for now.

garethellis0:
LGTM! :+1: :tada: 

(Just make sure to respond to all review comments in the future)

PR #320, opened 2018-05-13
Obstacle Manager: Part 2.5 (Created a geometry package)

garethellis0:
Created a package `sb_geom` that contains classes and utility functions for working with splines and polynomial lines. 

~~**NOTE:** The diff for this PR will become much nicer once all the dlib files are merged (#321)~~
**EDIT:** #321 has been merged, so the diff is good now.

PR #321, opened 2018-05-15
Obstacle Manager: Part 2 - Added the dlib library under src/sb_geom

garethellis0:
Because the system version of dlib in ubuntu 16 is slightly too old, I've add to add the library here to gain access to functions for global optimization. 

Not too much to review here honestly, but it should make reviewing #320 much easier, as the diff will only show the actually relevant files. Just need someone to approve once CI passes.


ValRat:
Might require disabling lockdown and re-enabling after, @garethellis0.

garethellis0:
I added a `.clang-format` file in `src/sb_geom/dlib` which should make clang format ignore that folder. @ValRat - will need re-appproval

PR #323, opened 2018-05-19
Pull request for node rename from pathfinding to twistcalculation

WilliamGu98:


ValRat:
I think you accidentally included the submodules back in, should remove those so CI stops complaining.

garethellis0:
Ok, went over this with the software leads, it doesn't seem like we should need an entirely new package just for this node. We think the simplest thing here would just be to rename the node to `twist_calculation` or `twist_calculator` and leave it in the `pathfinding_igvc` package, unless you forsee any issues with this? Sorry for all the confusion :disappointed:

PR #327, opened 2018-05-24
Path Finding: Part 3

mingyokim:
THE NODE!! Please only look at commits after 4dec7b5: Part 2 is not merged just yet


src/pathfinding_igvc/include/PathFinderNode.h
```diff
@@ -0,0 +1,48 @@
...
+    tf::TransformListener* _listener;
+
+    std::string _global_frame_name;
+    std::string _base_frame_name;
+
+    geometry_msgs::Point _goal;
+    nav_msgs::OccupancyGrid _grid;
+
+    bool _received_goal  = false;
+    bool _receivied_grid = false;
```
garethellis0: Just for consistency, we should probably put these under the `private:` line, which makes the fact that they're private more explicit.

src/pathfinding_igvc/include/PathFinderNode.h
```diff
@@ -0,0 +1,48 @@
...
+
+  public:
+    PathFinderNode(int argc, char** argv, std::string node_name);
+
+  private:
+    void occupancyGridCallback(const nav_msgs::OccupancyGrid grid);
+    void goalCallback(const geometry_msgs::Point goal);
+
+    void publishPath();
```
garethellis0: functions should have javadoc style comments

src/pathfinding_igvc/src/PathFinderNode.cpp
```diff
@@ -0,0 +1,79 @@
+/*
+ * Created By: Min Gyo Kim
+ * Created On: May 23rd 2018
+ * Description: Implementation of Path Finder Node
```
garethellis0: It might be helpful for people in the future if we could make this a bit more descriptive

src/pathfinding_igvc/src/PathFinderNode.cpp
```diff
@@ -0,0 +1,79 @@
...
+    SB_getParam(private_nh,
+                std::string("global_frame_name"),
+                this->_global_frame_name,
+                std::string("/map"));
+    SB_getParam(private_nh,
+                std::string("base_frame_name"),
+                this->_base_frame_name,
+                std::string("/base_link"));
+
+    std::string grid_subscriber_topic = "/occupancy_grid"; // dummy topic name
```
garethellis0: I get what you mean by `dummy topic name` here, but I think it could be a bit confusing, as most of our topics are "dummy" and get re-mapped, but we normally don't say that.

PR #328, opened 2018-09-02
Update From ROS Kinetic To ROS Melodic

garethellis0:
Still tackling a few teething issues, but mostly finished.


garethellis0:
Looks like the only thing left to fix is some pathfinding tests. Talked to @mingyokim and he'll take a look and we'll see what we come up with.

garethellis0:
Yeah, sadly the sick libraries haven't been updated since ROS indigo, so they still work but throw a crap ton of "deprecated" warnings.

PR #322, opened 2018-05-18
Obstacle Manager Part 3

garethellis0:
This contains the Obstacle Manager itself

PR #330, opened 2018-09-06
Obstacle Manager Part 3 - Added Obstacle Manager and Corresponding Tests

garethellis0:
This went through review in https://github.com/UBC-Snowbots/Snowflake/pull/322, but we made a bunch of changes at competition, and I've removed a few things to make it _slightly_ smaller.

There are still many `TODO`'s and `NOTE`'s in the code, but at this point, since we're not sure how much (if any) of this we're going to be using for URC, I don't think it's a good idea to put more engineering time into fixing it up until we decide what we actually need.


garethellis0:
Currently waiting on #330 to get merged as my computer is now running Ubuntu 18.04.

PR #350, opened 2018-09-22
Fix path constructor melodic

mingyokim:
Uncommenting broken tests from #328 

Issue was that the pointer to `OccupancyGridAdapter` in `PathConstructor` got corrupted. So I changed it so that `PathConstructor` takes in a pointer of `OccupancyGridAdapter` instead of an instance/object of it and storing its reference. This would force whoever is calling `PathConstructor` to deal with `OccupancyGridAdapter`'s pointer.

### DIFF
--- a/src/pathfinding_igvc/include/PathConstructor.h
+++ b/src/pathfinding_igvc/include/PathConstructor.h
@@ -26,7 +26,7 @@ class PathConstructor {
      * @param occupancy_grid_adapter
      * @return PathConstructor
      */
-    PathConstructor(OccupancyGridAdapter occupancy_grid_adapter);
+    PathConstructor(OccupancyGridAdapter *occupancy_grid_adapter);
 
     /**
      * Takes in a stack of GridPoints and returns a path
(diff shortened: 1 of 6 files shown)


src/pathfinding_igvc/include/PathConstructor.h
```diff
@@ -13,7 +13,7 @@
 #include <OccupancyGridAdapter.h>
 
 class PathConstructor {
-    OccupancyGridAdapter* _occupancy_grid_adapter;
+    std::shared_ptr<OccupancyGridAdapter> _occupancy_grid_adapter;
```
mingyokim: My first time using this, but it seems to fit this case since the function creating `PathConstructor` would be creating a pointer to a certain `OccupancyGridAdapter` instance, and copying the pointer to here. And the shared pointer would automatically get destroyed after all references to the instance are gone.

https://en.cppreference.com/w/cpp/memory/shared_ptr

```
The object is destroyed and its memory deallocated when either of the following happens:

the last remaining shared_ptr owning the object is destroyed;
the last remaining shared_ptr owning the object is assigned another pointer via operator= or reset().
```

garethellis0:
Whoops, sorry, I accidentally merged in `master` to this branch. Hopefully shouldn't cause issues, but maybe just double check to be sure.

mingyokim:
@garethellis0 the merge commit only has changes on README, so it shouldn't affect anything

PR #351, opened 2018-09-29
Added rgbd_launch as a rosdep in sb_pointcloud_processing

garethellis0:


setup_scripts/install_dependencies.sh
```diff
@@ -40,6 +40,7 @@ cd $CURR_DIR
 sudo rosdep init
 # Update Rosdeps
 rosdep update
+
```
mingyokim: don't we need to add a line here for installing the package, like `sudo apt-get install ros-melodic-rgbd_launch`? Or does `build_depend` and `run_depend` install it for you?

garethellis0: Adding a `build_depend` or `run_depend` line will cause them to be installed when we run the `rosdep --install` command below.

PR #355, opened 2018-10-13
Ball extractor

mingyokim:
## Please fill out the following before requesting review on this PR

**Description of Change:** 
Created `BallExtractorNode`, which is almost identical to `LineExtractorNode`. Didn't do any refactoring and just copy pasted `LineExtractorNode` with necessary modifications, so there are quite a bit of duplicate functions between them.

Overview of `BallExtractorNode`:

1. Receives a point cloud filtered to roughly only contain the points representing the tennis ball (after HSV/height filter)
2. Performs DBSCAN to cluster the points in the YZ plane. With `LineExtractor`, since the lanes were on the ground we were using the XY plane/birds eye view. But with ball extractor, the ball is "in front" of the robot, so it makes sense to use the YZ plane. Also, ideally it only returns one cluster for the tennis ball, but currently it just expects that the first cluster contains the tennis ball. So if we often get more than one clusters, we could just pick the one with the most points or something.
3. Estimates the center of the tennis ball based on the cluster. Currently, I just calculate the min and max of y and z values in the cluster, and calculate the average of them. And I just set x as the minimum value of x in the cluster. These should do the job for now, but can modify it later if we find problems with it.

### Checklist
- [x] **Start of document comments**: each `.cpp` and `.h` file should have a comment at the start of it. See files in `src/sample_package` for examples.
- [x] **Function comments**: All function definitions (usually in the `.h` file) should have a javadoc style comment at the start of them. For examples, see the classes defined in `src/sample_package`
- [x] **Remove all commented out code**
- [x] **Remove extra print statements**: for example, those just used for testing
- [x] **Resolve all TODO's**: All `TODO` (or similar) statements should either be completed or associated with a github issue


garethellis0:
@mingyokim - just double checking, is this ready for review?

mingyokim:
Nope, I’ll let you know when it is

mingyokim:
@garethellis0 ready for review !

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+ * Created By: Min Gyo Kim
+ * Created On: September 29, 2018
+ * Description: Header file for Ball Extractor Node
+ *  This node subscribes to a point cloud topic and expects the topic to contain
+ * a point cloud
+ *  that is filtered to contain roughly only the points that represent the
+ * tennis ball.
+ *  Then the node performs DBSCAN on the Y-Z plane, and calculates a rough
+ * estimate of the
+ *  center of the tennis ball, and publishes the point.
```
garethellis0: looks like clang messed up this comment a bit, we should re-format it.

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+#include <sb_utils.h>
+#include <sensor_msgs/PointCloud2.h>
+#include <string>
+
+class BallExtractorNode {
+  public:
+    /*
+     * @clusters stores the output from DBSCAN
+     */
+    std::vector<pcl::PointCloud<pcl::PointXYZ>> clusters;
```
garethellis0: Member variables should be private and should be accessible only via getters. Here's a few reasons why this is important: https://stackoverflow.com/questions/1568091/why-use-getters-and-setters-accessors

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+    void extractBall();
+
+    /*
+     * @clusters: input clusters
+     * @cluster_points: vector to store converted cluster points
+     * @colors: vector to store color for each cluster point
+     * This function takes in @clusters, assigns a different color
+     * for each cluster, and converts each point to geometry_msgs::Point.
+     * It pushes the points and colors to the back of @cluster_points
+     * and @colors.
```
garethellis0: Funky formatting here, we should re-format this into a proper javadoc comment, ie.
```
/**
 * One line description here
 * 
 * More through description here, possibly spanning over multiple 
 * lines
 *
 * @param some_param This is a parameter that we use for things
 * @return Some function of the parameter
 */
```

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+     * @pclPtr stores the pointer to the PCL PointCloud after it has
+     * been converted from sensor_msgs PointCloud2
+     */
+    pcl::PointCloud<pcl::PointXYZ>::Ptr pclPtr;
+
+    /*
+     * The callback function is called whenever the node receives a
+     * PointCloud message. It converts sensor_msgs PointCloud2 pointer
+     * to PCL PointCloud pointer and then extracts lines from the PointCloud
+     */
```
garethellis0: missing `@param`

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+     * with a different color for each cluster and publishes a message
+     * to rviz for visualization at "~/debug/clusters".
+     */
+    void visualizeClusters();
+
+    /*
+     * This function makes a Marker for the given geometry_msgs::Point
+     * and publishes it to rviz for visualization at "~/debug/ball".
+     */
+    void visualizeBall(geometry_msgs::Point ball);
```
garethellis0: would be good to see the `visualizeX` methods have the same style of interface (ie. passing in what to visualize as arguments or just not taking any parameters). I'd advocate for both taking what they're going to visualize as a parameter, makes it easier to debug/test.

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+
+    /*
+     * Get the center of the cluster
+     * The z value of the center is min and max / 2
+     * The y value of the center is min and max / 2
+     * The x value of the center is the min
+     * @cluster_index: the index of cluster of interest in @clusters
+     */
+    geometry_msgs::Point getCenterOfCluster(unsigned int cluster_index);
```
garethellis0: Could we just pass in the cluster by const reference (ie. `const pcl::PointCloud<pcl::PointXYZ>&`)? Seems like it would be less error-prone (can't just pass in a random index and get an "out-of-bounds" error. Then we could also maybe make this `static`?

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -16,6 +16,10 @@ using namespace std;
 using namespace std::tr1;
 
 class DBSCAN {
+  public:
+    enum Plane { XY, YZ };
```
garethellis0: Does `DBSCAN` only work in 2D? Quick doc comment for this enum would be helpful.

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -46,17 +50,25 @@ class DBSCAN {
      */
     vector<unsigned int>* _neighbors;
 
-    int _min_neighbors = 5;
-    float _radius      = 5;
-
+    int _min_neighbors               = 5;
+    float _radius                    = 5;
     unsigned int _sequential_cut_off = 1000;
```
garethellis0: In general, we should avoid initializing member variables like this, and should instead use constructor initialization lists: https://www.cprogramming.com/tutorial/initialization-lists-c++.html

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -46,17 +50,25 @@ class DBSCAN {
...
+    Plane _plane;
 
-  public:
     /*
      * Constructor:
      * Takes in minimum number of neighbours and radius as parameters
+     * as well as plane (XY or YZ)
```
garethellis0: We should remove `(XY or YZ)` here in case the enum ever changes (because it's likely this doc comment wouldn't get updated along with the enum).

src/sb_pointcloud_processing/include/DBSCAN.h
```diff
@@ -46,17 +50,25 @@ class DBSCAN {
...
-    DBSCAN(int min_neighbours = 5, float radius = 5);
+    DBSCAN(int min_neighbours = 5, float radius = 5, Plane plane = XY);
+    ~DBSCAN() { delete[] this->_neighbors; }
+
+  public:
+    static vector<pcl::PointCloud<pcl::PointXYZ>>
+    getClusters(pcl::PointCloud<pcl::PointXYZ>::Ptr pcl_ptr,
+                int min_neighbours = 5,
+                float radius       = 5,
+                Plane plane        = XY);
```
garethellis0: I'm ok with setting default values in the constructor, but I'd rather we explicitly pass them instead of also having default values for method arguments. This helps prevent errors like "that's weird, I changed `min_neighbours` in my call to the constructor, but it's not doing anything....".

src/sb_pointcloud_processing/include/BallExtractorNode.h
```diff
@@ -0,0 +1,134 @@
...
+     * Checks whether or not all the params we are getting from NodeHandler are
+     * valid
+     * params being checked: degree_polynomial, lambda, min_neighbours, radius
+     */
+    bool areParamsInvalid();
+
+    /*
+     * Number of point clouds processed so far
+     */
+    int _execution_count = 0;
```
garethellis0: I think we can remove this once we get remove the performance debug logging below?

src/sb_pointcloud_processing/scripts/kill_pcl.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
garethellis0: Quick doc comment explaining what this script does, why it's here, and why it's necessary would be helpful.

src/sb_pointcloud_processing/src/BallExtractorNode.cpp
```diff
@@ -0,0 +1,221 @@
...
+    // store converted pointcloud for use
+    this->pclPtr = temp_cloud;
+
+    // extract ball from the pointcloud
+    extractBall();
+
+    if (++this->_execution_count % 100 == 0) {
+        std::cout << this->_execution_count << ": " << ros::Time::now().toNSec()
+                  << endl;
+    }
```
garethellis0: We should remove the performance logging here.

src/sb_pointcloud_processing/src/BallExtractorNode.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+    std::string ns = "debug";
+
+    visualization_msgs::Marker marker =
+    snowbots::RvizUtils::createMarker(ball, colors, scale, this->frame_id, ns);
+
+    rviz_ball_publisher.publish(marker);
+}
+
+// TODO: duplicate function from LineExtractorNode - refactor this
```
garethellis0: TODO?

src/sb_pointcloud_processing/src/BallExtractorNode.cpp
```diff
@@ -0,0 +1,221 @@
...
+    rviz_cluster_publisher.publish(marker);
+}
+
+void BallExtractorNode::convertClustersToPointsWithColors(
+std::vector<pcl::PointCloud<pcl::PointXYZ>> clusters,
+std::vector<geometry_msgs::Point>& cluster_points,
+std::vector<std_msgs::ColorRGBA>& colors) {
+    std::vector<float> color_library_r = {1.0, 0.0, 0.0};
+    std::vector<float> color_library_g = {0.0, 0.0, 1.0};
+    std::vector<float> color_library_b = {0.0, 1.0, 0.0};
```
garethellis0: If these are fixed size, we should probably use `std::array`

src/sb_pointcloud_processing/src/BallExtractorNode.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+            geometry_msgs::Point msg_point;
+            msg_point.x = pcl_point.x;
+            msg_point.y = pcl_point.y;
+
+            cluster_points.push_back(msg_point);
+            colors.push_back(color);
+        }
+    }
+}
```
garethellis0: @RobynCastro - does this sort of stuff belong in RvizUtils?

src/sb_pointcloud_processing/src/BallExtractorNode.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+geometry_msgs::Point
+BallExtractorNode::getCenterOfCluster(unsigned int cluster_index) {
+    pcl::PointCloud<pcl::PointXYZ> cluster = this->clusters[cluster_index];
+
+    struct {
+        bool operator()(pcl::PointXYZ a, pcl::PointXYZ b) const {
+            return a.y < b.y;
+        }
+    } lessY;
```
garethellis0: Could simplify this a bit to:
```
auto lessY = [&](const pcl::PointXYZ& a, const pcl::PointXYZ& b) {return a.y < b.y; };
```
(I think)
(also this way we're passing by const reference instead of by value, so we don't copy each point)
(same sort of thing could be applied below as well)

src/sb_pointcloud_processing/src/DBSCAN.cpp
```diff
@@ -109,9 +118,18 @@ void DBSCAN::findNeighbors() {
...
+
+    switch (this->_plane) {
+        case DBSCAN::YZ:
+            dy = abs(p1.y - p2.y);
+            dz = abs(p1.z - p2.z);
+            return sqrt(pow(dy, 2) + pow(dz, 2));
+        default:
+            dx = abs(p1.x - p2.x);
+            dy = abs(p1.y - p2.y);
+            return sqrt(pow(dx, 2) + pow(dy, 2));
```
garethellis0: We should probably change the `default` case to `DBSAN:XY` here, and make `default` throw an exception, because it means we were given a plane that we don't know how to handle.

src/sb_pointcloud_processing/test/ball_extractor_rostest.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    EXPECT_FLOAT_EQ(x, ballObstacle.x);
+    float y_min = *std::min_element(ys.begin(), ys.end());
+    float y_max = *std::max_element(ys.begin(), ys.end());
+    EXPECT_FLOAT_EQ((y_min + y_max) / 2, ballObstacle.y);
+
+    float z_min = *std::min_element(zs.begin(), zs.end());
+    float z_max = *std::max_element(zs.begin(), zs.end());
+    EXPECT_FLOAT_EQ((z_min + z_max) / 2, ballObstacle.z);
+}
```
garethellis0: Super cool test!

garethellis0:
@mingyokim - looks like there are still some comments here?

PR #365, opened 2018-11-17
Added Rviz Polygon Markers

RobynCastro:
### Description
- Added support for polygon markers in Rviz

### Testing Done
- Ran the rostests and it succeeds all tests.

### Resolved Issues
N/A

### Review Checklist
***It is the reviewers responsibility to also make sure every item here has been covered***
- [X] **Start of document comments**: each `.cpp` and `.h` file should have a comment at the start of it. See files in `src/sample_package` for examples.
- [X] **Function comments**: All function definitions (usually in the `.h` file) should have a javadoc style comment at the start of them. For examples, see the classes defined in `src/sample_package`
- [X] **Remove all commented out code**
- [X] **Remove extra print statements**: for example, those just used for testing
- [X] **Resolve all TODO's**: All `TODO` (or similar) statements should either be completed or associated with a github issue


src/sb_utils/test/rviz_utils_rostest.cpp
```diff
@@ -0,0 +1,211 @@
+#include "RvizUtils.h"
+#include <gtest/gtest.h>
+
```
garethellis0: Quick comment on why this is a RosTest (I presume because of timestamps and whatnot) would be good here

src/sb_utils/include/RvizUtils.h
```diff
@@ -73,6 +77,97 @@ class RvizUtils {
...
+     */
+    template <typename T>
+    visualization_msgs::Marker
+    createPolygonMarker2D(T polygon,
+                          visualization_msgs::Marker::_color_type color,
+                          visualization_msgs::Marker::_scale_type scale,
+                          std::string frame_id,
+                          std::string ns,
+                          int type = visualization_msgs::Marker::LINE_STRIP,
+                          int id   = 1) {
```
RobynCastro: I don't know the proper conventions for template functions, so I am not sure how to make it look good.

garethellis0: Why are we templating here?

RobynCastro: At the time I thought it'd be useful when working with any type of point. But thinking about it now we should always be using the snowbots specific points. I will change it back.

garethellis0:
Just one comment left to resolve (and update the branch) and this should be good to go!

garethellis0:
CI build failed because of a 10m timeout, so I restarted it.

.gitignore
```diff
@@ -22,6 +22,9 @@ cmake-build-release
 
 .rosinstall.bak
 
+# 3rd Party Libraries
+external_libs
```
garethellis0: if this works but the below one doesn't (line 30), we should get rid of the old one

src/sb_utils/src/RvizUtils.cpp
```diff
@@ -6,20 +6,40 @@
...
+
+namespace snowbots {
+namespace RvizUtils {
+    /**
+     * Helper function that sets up common marker parameters
+     *
+     * @param scale the scale
+     * @param frame_id the frame id
+     * @param ns the namespace
+     * @param type the type of marker
```
garethellis0: this is just a number, could we specify that it's from `visualization_msgs::Marker::*`? Same goes if we pass the type like this in other places here.

src/sb_utils/src/RvizUtils.cpp
```diff
@@ -98,7 +205,11 @@ RvizUtils::createMarkerColor(float r, float g, float b, float a) {
     return color;
 }
 
-Marker::_scale_type RvizUtils::createrMarkerScale(float x, float y, float z) {
+// visualization_msgs::Marker::_scale_type
+// createMarkerScale(float x, float y, float z);
```
garethellis0: Could we remove this?

src/sb_utils/test/rviz_utils_rostest.cpp
garethellis0: bump?

src/sb_utils/test/rviz_utils_rostest.test
```diff
@@ -0,0 +1,3 @@
+<launch>
+    <test test-name="rviz_utils_rostest" pkg="sb_utils" type="rviz_utils_rostest" />
+</launch>
```
garethellis0: nit; github complaining about end-of-file here

src/sb_utils/test/rviz_utils_rostest.cpp
RobynCastro: Should already be done, it should be at the top of the file, not sure why it's not showing up here.

garethellis0:
Only one comment left. CI seems to be timing out here when running the tests. Can you confirm that when you run this branch locally all the tests run a pass fine? 

My _guess_ is that there is a ROSTest that never terminates, but I'm just shooting in the dark.

PR #371, opened 2019-02-15
Moved PCL Transformation into the sb_pointcloud_processing package

RobynCastro:
## Description
Changed zed_transform into a generic pointcloud_transform.

### Testing Done
Used rviz to observe pointcloud transforms.

### Resolved Issues
N/A

### Review Checklist
***It is the reviewers responsibility to also make sure every item here has been covered***
- [x] **Start of document comments**: each `.cpp` and `.h` file should have a comment at the start of it. See files in `src/sample_package` for examples.
- [x] **Function comments**: All function definitions (usually in the `.h` file) should have a javadoc style comment at the start of them. For examples, see the classes defined in `src/sample_package`
- [x] **Remove all commented out code**
- [x] **Remove extra print statements**: for example, those just used for testing
- [x] **Resolve all TODO's**: All `TODO` (or similar) statements should either be completed or associated with a github issue

### DIFF
--- a/src/sb_pointcloud_processing/CMakeLists.txt
+++ b/src/sb_pointcloud_processing/CMakeLists.txt
@@ -63,65 +63,65 @@ include_directories(
 ## With catkin_make all packages are built within a single CMake context
 ## The recommended prefix ensures that target names across packages don't collide
 add_executable(line_extractor_node
-  include/LineExtractorNode.h
-  include/DBSCAN.h
-  include/Regression.h
-  src/line_extractor_node.cpp
-  src/LineExtractorNode.cpp
-  src/DBSCAN.cpp
-  src/Regression.cpp
-  include/Regression.h
+    include/LineExtractorNode.h
+    include/DBSCAN.h
+    include/Regression.h
+    src/line_extractor_node.cpp
+    src/LineExtractorNode.cpp
+    src/DBSCAN.cpp
+    src/Regression.cpp
+    include/Regression.h
 )
 
-add_library(sb_pointcloud_processing
-  include/rgb_to_hsv.h
-  src/rgb_to_hsv.cpp
-  include/ColourspaceConverter.h
-  src/ColourspaceConverter.cpp
-)
-
-add_executable(test_pcl_generator_node
-  src/test_pcl_generator_node.cpp
-  test/TestUtils.h
-)
-
-add_executable(pcl_transform src/pcl_transform.cpp)
-
-## Specify libraries to link a library or executable target against
 target_link_libraries(line_extractor_node
-  ${catkin_LIBRARIES}
-  ${PCL_COMMON_LIBRARIES}
-  ${PCL_IO_LIBRARIES}
+    ${catkin_LIBRARIES}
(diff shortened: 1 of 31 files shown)


WilliamGu98:
I don't have time right now, but I should be able to finish a review by tomorrow night.

garethellis0:
Ok, I'm gonna let @WilliamGu98 take the first pass on this one!

src/sb_pointcloud_processing/CMakeLists.txt
```diff
@@ -69,47 +74,55 @@ add_executable(line_extractor_node
...
   ${catkin_LIBRARIES}
   ${PCL_COMMON_LIBRARIES}
   ${PCL_IO_LIBRARIES}
-  )
+)
+
+target_link_libraries(pcl_transform
+  ${catkin_LIBRARIES}
+  ${sb_utils_LIBRARIES}
+)
```
garethellis0: out-of-scope (so feel free to ignore); but could we put the `target_link_libraries` calls for `test_pcl_generator_node` and `pcl_transform` directly below their respective `add_executable` calls above? I think that will make this a bit easier to follow.

src/sb_pointcloud_processing/src/pcl_transform.cpp
```diff
@@ -14,38 +14,56 @@
 #include <sb_utils.h>
 
 ros::Publisher pub;
-std::string output_frame;
 tf2_ros::Buffer tf_buffer;
 
+std::string output_frame;
+
+// The amount of time (s) for a new pointcloud transform to be published.
```
garethellis0:
``` suggestion
// The amount of time(s) for a new pointcloud transform to be published.
````

File: src/sb_pointcloud_processing/src/pcl_transform.cpp [hunk 3]
```diff
@@ -14,38 +14,56 @@
...
+        output.header.frame_id = output_frame;
+
         // Publish the transformed pointcloud
         pub.publish(output);
-    } catch (tf2::TransformException ex){
+
+        ros::Duration(transform_period).sleep();
+    } catch (tf2::TransformException ex) {
         ROS_WARN("%s", ex.what());
         ros::Duration(1.0).sleep();
```
garethellis0: Hmm, does it make sense to sleep here? That means if there is a TF issue on one frame, we won't transform anything for 1 second? I think this is here to make sure we don't overwhelm the log with messages, so perhaps a timer-based system so that we only _log_ every 500ms or so might be a better option?

RobynCastro: What do you mean by this? Do you mean to change the sleep duration to 0.5s? Or do you mean create a timer callback that logs the status of the system every 0.5s?

garethellis0: I mean remove the `sleep` here so that we're still trying to transform things as fast as possible, but add a check to see if >1s has passed since the last time we logged so that we don't overwhelm the log by logging 30 times a second (or whatever the framerate is)

PR #380, opened 2019-10-19
Merge README for new members into README

raadkhan:


raadkhan:
this is a PR from `master` to `master` - bad practice so don't do this

PR #382, opened 2020-03-07
Motor control

ihsan314:
### Description

This node, given a twist message, will control the motors using the phidgets libraries. For now, the dimensions of the actual robot aren't known, so it currently controls a single motor, but eventually all motors can be controlled in the manner shown in this node.

### Testing Done

For the time being, I did experimental tests on one motor, to make sure it was properly activated upon receipt of a twist message.


src/motor_control_circ/package.xml
```diff
@@ -0,0 +1,65 @@
...
+  <version>0.0.0</version>
+  <description>The motor_control package</description>
+
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="ihsan@todo.todo">ihsan</maintainer>
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
```
raadkhan: ...same for with these comments

src/motor_control_circ/src/MoveMotor.cpp
```diff
@@ -0,0 +1,37 @@
...
+    ROS_INFO("Received Twist");
+
+    PhidgetBLDCMotorHandle bldcMotor0;
+
+    PhidgetBLDCMotor_create(&bldcMotor0);
+
+    int channel = 2;
+    Phidget_setHubPort((PhidgetHandle)bldcMotor0, channel);
+
+    Phidget_openWaitForAttachment((PhidgetHandle)bldcMotor0, 5000);
```
raadkhan: what happens if the motor is not detected for a long enough time?

src/motor_control_circ/src/MoveMotor.cpp
```diff
@@ -0,0 +1,37 @@
...
+    ros::NodeHandle nh;
+    ros::NodeHandle private_nh("~");
+
+    std::string subscribe_topic = "subscribe_topic";
+    int queue_size = 10;
+    my_subscriber = nh.subscribe(subscribe_topic, queue_size, &MoveMotor::callback, this);
+}
+
+void MoveMotor::callback(const geometry_msgs::Twist::ConstPtr& msg) {
+    ROS_INFO("Received Twist");
```
raadkhan: let's print out the value of the twist as well

src/motor_control_circ/src/MoveMotor.cpp
```diff
@@ -0,0 +1,37 @@
...
+    PhidgetBLDCMotor_create(&bldcMotor0);
+
+    int channel = 2;
+    Phidget_setHubPort((PhidgetHandle)bldcMotor0, channel);
+
+    Phidget_openWaitForAttachment((PhidgetHandle)bldcMotor0, 5000);
+
+    PhidgetBLDCMotor_setTargetVelocity(bldcMotor0, msg->linear.x);
+
+    usleep(5 * 1000 * 1000);
```
raadkhan: is this for testing that the motor moves according to `msg->linear.x` for 5 seconds?

src/motor_control_circ/src/MoveMotor.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    Phidget_close((PhidgetHandle)bldcMotor0);
+    PhidgetBLDCMotor_delete(&bldcMotor0);
```
raadkhan: nice clean up here

ihsan314:
Thanks for the review. I'll apply the fixes when I get the chance.

src/motor_control_circ/src/MoveMotor.cpp
ihsan314: Yes.

src/motor_control_circ/src/MoveMotor.cpp
ihsan314: I believe that's only the case when the motor is not actually connected, i.e. there's nothing to attach to.

ihsan314:
The build system fairly frustrating to work with. Although due to the .travis.yml additions, the library is installed, somehow the header files are still not visible in the build. It might have to do with the fact that the docker used to do the compilation somehow isolates the compilation process from everything else. I will take a closer look later.

PR #391, opened 2020-10-15
Fix of installation script

ihsan314:
### Description
The [ROS installation instructions](https://wiki.ros.org/melodic/Installation/Ubuntu) have changed for ROS Melodic, as `python-rosdep` and `python-rosinstall-generator` are no longer installed by default when installing `ros-melodic-desktop-full`. I added the explicit installation of the two packages.

The history is a bit messed up because the master branch on my fork is messed up. I would fix it, but this branch is only 1 addition/deletion ahead of the Snowbots master branch, and will get squashed anyway. I'll deal with the master branch of my fork later.


kevinlinxc (approved):
Looks good. CI passed.

raadkhan (approved):
Also LGTM

PR #402, opened 2021-08-01
Dual cameras

kevinlinxc:
### Description
Adds launch files for the two intel cameras, and makes the installation of the intel realsense sdk and ros wrapper automatically the highest version. 

### Testing Done
Running `roslaunch cameras realsense_d435i.launch` streams our d435i camera to a rqt_image_view window automatically
Running `roslaunch cameras realsense_d415.launch` does the same
Running `roslaunch cameras realsense_both.launch` creates two rqt_image_view windows and streams both cameras.

If any number of cameras aren't plugged in, the node doesn't crash, rather it does nothing until a camera is plugged in.

### DIFF
--- a/.rosinstall
+++ b/.rosinstall
@@ -2,13 +2,11 @@
     local-name: src/external_pkgs/nmea_navsat_driver
     uri: https://github.com/UBC-Snowbots/nmea_navsat_driver.git
 - git:
-    local-name: src/external_pkgs/realsense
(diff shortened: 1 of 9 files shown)


ihsan314 (approved):
Nice job, this is ready to be merged in.

ihsan314 (approved):
You should be able to squash and merge now.

